PR #8, opened 2019-02-19
Adds article group into the blog card output

b-m-f:
# Done
In order to display the correct styling for the different article groups we need to pass the group through to the template.
This PR resolves the group name for any given article and then applies those in the template.

# QA

Here is a screenshot of the updated plugin used on jp.ubuntu.com
image

### DIFF
--- a/canonicalwebteam/blog/logic.py
+++ b/canonicalwebteam/blog/logic.py
@@ -52,7 +52,7 @@ def replace_images_with_cloudinary(content):
 def transform_article(
     article, featured_image=None, author=None, optimise_images=False
 ):
-    """Transform article to include featured image, human readable
+    """Transform article to include featured image, a group, human readable
     date and a stipped version of the excerpt
 
     :param article: The raw article object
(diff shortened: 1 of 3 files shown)


tbille:
@b-m-f I don't mind merging this, but we should be careful when adding api calls like this. Even if we cache the API calls it still adds more http call to the view, which is already quite slow when loading for the first time.

And if we merge this, can you update the version so that we can publish it to pypi.

anasereijo:
@b-m-f this is looking so much better, thanks!

Just a couple of things:

- [ ] author should be a link 

- [ ] the titles should all be horizontally aligned (see example below)
ubuntu blog 1

please check [blog.ubuntu](https://blog.ubuntu.com/) for reference

b-m-f:
@toto, this is true. I was thinking to fetch all of those with http/2. Might be an enhancement for the http module

toto:
Not the toto you are thinking of :-D

tbille:
@b-m-f b-m-f this is a different

tbille:
@b-m-f 

> @tbille, this is true. I was thinking to fetch all of those with http/2. Might be an enhancement for the http module

This is something different. I think you can keep this code for now. We have some fun improvements that we could bring later on the project.

PR #10, opened 2019-03-07
Creates and exports django app on blog module

b-m-f:
# Done
Fixes https://github.com/ubuntudesign/base-squad/issues/385

Creates a skeleton Django app to export on the blog module

# QA
## Django
- Create a new Django App with `django-admin startproject qatest .`
- in the `qatest/urls.py` file of the qatest app use the following code
```
from django.urls import path, include
urlpatterns = [path(r"blog/", include("canonicalwebteam.blog.django.urls"))]
```

- run the django app with `python manage.py runserver`
- going to `http:localhost:8000/blog/hi` should return `{"route": "index"}` JSON from the Django blog app

## Flask
- Clone https://github.com/canonical-websites/jp.ubuntu.com
- copy `canonicalwebteam` folder into the cloned jp.ubuntu.com folder
- remove `canonicalwebteam.blog` from `requirements.txt`
- `./run` jp.ubuntu.com and make sure the blog still displays on the site


tbille:
There is no circle ci intergrated here yet. Can you run `flake8` before pushing?

```
$ flake8 canonicalwebteam
canonicalwebteam/blog/logic.py:41:9: E999 SyntaxError: invalid syntax
canonicalwebteam/blog/app.py:2:1: F401 'canonicalwebteam.blog.djangoblog.apps.DjangoBlogConfig' imported but unused
canonicalwebteam/blog/djangoblog/views.py:4:1: F401 'django.shortcuts.render' imported but unused
canonicalwebteam/blog/djangoblog/tests.py:4:1: F401 'django.test.TestCase' imported but unused
```

canonicalwebteam/blog/djangoblog/apps.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
tbille: This is used if we want python 2 support. Is it something we need? (@nottrobin )

canonicalwebteam/blog/flaskblog/views.py
```diff
@@ -1,10 +1,10 @@
-import flask
-
 from canonicalwebteam.blog import wordpress_api as api
 from canonicalwebteam.blog import logic
 
 
 def build_blueprint(blog_title, tags_id, tag_name):
+    import flask
```
tbille: is there a reason why you added this here?

canonicalwebteam/blog/djangoblog/tests.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.test import TestCase
+
+# Create your tests here.
```
tbille: I don't think you need to have the tests file if no tests (even if it's just a bootstrap)

And the pattern we have for tests is to have a `tests` at root of the project. We could change this (more js style) but for now I would prefer to keep our patterns.

canonicalwebteam/blog/app.py
```diff
@@ -1,4 +1,5 @@
-from canonicalwebteam.blog.views import build_blueprint
+from canonicalwebteam.blog.flaskblog.views import build_blueprint as flaskblog
```
tbille: please don't rename expect if you have 2 imports with same name.

canonicalwebteam/blog/flaskblog/views.py
b-m-f: Have updated it. The dynamic import now happens in the main `app.py`. I think that way is cleaner.

The original reason was, to not having to install flask in a django app

b-m-f:
So, I am having some trouble with the flake8 here.
This line `canonicalwebteam/blog/logic.py:41:9: E999 SyntaxError: invalid syntax` is the problem. It does get autofixed by black into a state that seems to throw of flake8. Any idea @tbille ?

tbille:
@b-m-f can you run `flake8` with python 3and see if you still get the same error message?

canonicalwebteam/blog/flaskblog/views.py
```diff
@@ -1,5 +1,4 @@
 import flask
-
```
tbille: keep line

canonicalwebteam/blog/app.py
```diff
@@ -1,4 +1,4 @@
-from canonicalwebteam.blog.views import build_blueprint
+from canonicalwebteam.blog.djangoblog.apps import DjangoBlogConfig  # noqa: 401
```
tbille: shoudln''t be there

canonicalwebteam/blog/app.py
```diff
@@ -15,5 +12,7 @@ def __init__(
             self.init_app(app, blog_title, tag_id, tag_name, url_prefix)
 
     def init_app(self, app, blog_title, tag_id, tag_name, url_prefix):
+        from canonicalwebteam.blog.flaskblog.views import build_blueprint
```
tbille: move at top of the file

canonicalwebteam/blog/djangoblog/apps.py
tbille: I forgot we already use f string. So Python 3

Please remove those lines

b-m-f: Updated :)

canonicalwebteam/blog/djangoblog/views.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
tbille: remove lines

b-m-f: thanks!

tbille:
can you add a small qa step to play a bit around with it?

tbille:
@b-m-f 	QA steps still don't work for me. Please add more details.

```
from django.urls import path
urlpatterns = [path(r"blog/", include("canonicalwebteam.blog.django.urls"))]

Missing include. I guess it comes from from django.urls import include
```

> run the django app with python manage.py runserver in the qatest folder

This is not generated in the qa test folder but at the root where you run the comand given

> going to http:localhost:8000/blog/hi should return {hi: "test"} JSON from the Django blog app

Not getting the same result: 

```
$ curl http://127.0.0.1:8000/blog/hi
{"route": "index"}
```

Haven't tested flask yet.

b-m-f:
The return is correct :)  Went out of sync cause of backporting things from https://github.com/canonical-webteam/blog-extension/pull/11 .

PR #13, opened 2019-03-19
Allow excluding tags

b-m-f:
# Done
Extends api to fetch all except excluded tags according to https://developer.wordpress.org/rest-api/reference/posts/

# QA
- The test should reflect the URL building correctly. Run with `python -m unittest tests/test_wordpress_api.py`

### DIFF
--- a/canonicalwebteam/blog/wordpress_api.py
+++ b/canonicalwebteam/blog/wordpress_api.py
@@ -18,26 +18,30 @@ def process_response(response):
     return response.json()
 
 
-def get_articles(tags, per_page=12, page=1, exclude=None, category=None):
-    url_parts = [
-        API_URL,
-        "/posts?",
-        "&per_page=",
-        str(per_page),
-        "&page=",
-        str(page),
-    ]
-
-    if tags:
-        url_parts = url_parts + ["&tags=", str(tags)]
-
-    if exclude:
-        url_parts = url_parts + ["&exclude=", str(exclude)]
-
-    if category:
-        url_parts = url_parts + ["&categories=", str(category)]
-
-    url = "".join(url_parts)
+def get_articles(
+    tags=None,
+    per_page=12,
+    page=1,
+    tags_exclude=[],
+    exclude=None,
+    categories=None,
+):
+    """
+    Get articles from Wordpress api
+    :param tags: Comma separated string of tags to fetch articles for
+    :param per_page: Articles to get per page
+    :param page: Page number to get
+    :param tags_exclude: Array of IDs of tags that will be excluded
+    :param exclude: Comma separated string of article IDs to be excluded
+    :param category: Comma separated list of categories, which articles
+        should be fetched
+    """
+    url = (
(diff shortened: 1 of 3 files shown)


canonicalwebteam/blog/wordpress_api.py
```diff
@@ -43,16 +43,16 @@ def get_articles(tags, per_page=12, page=1, exclude=None, category=None):
...
-    url = "".join(
-        [
-            API_URL,
-            "/posts?slug=",
-            slug,
-            "&tags=",
-            "".join(str(tag) for tag in tags),
-        ]
-    )
+def get_article(slug, tags=None, excluded_tags=None):
```
jpmartinspt: Is there any a case when you want tags and excluded_tags at the same time? If you pass both arguments it will output a URL with tags and excluded_tags.

b-m-f: Hm, good question. I dont think they need to be mutually exclusive since some articles may have multiple tags?
I would prefer leaving the possibility to add both.

jpmartinspt: Should the tests cover that case then?

b-m-f: Good idea! Will implement

nottrobin:
Fixes https://github.com/canonical-webteam/blog-extension/issues/12

canonicalwebteam/blog/wordpress_api.py
nottrobin: Actually the API will happily accept empty values for both `tags` and `exclude_tags` - https://admin.insights.ubuntu.com/wp-json/wp/v2/posts?tag=&exclude_tags= - which opens up the opportunity for the code to be significantly simpler.

Also, if the argument to the API is called "exclude_tags", that also seems descriptive enough for our use here. To use a very similar but slightly different variable name "excluded_tags" seems likely to cause confusion.

What do you think of:

``` python3
def get_article(slug, tags=[], exclude_tags=[]):
    url = (
        f"{API_URL}/posts?slug={slug}&tags={','.join(tags)}"
        f"&exclude_tags={','.join(exclude_tags)}"
    )
```

b-m-f: Have updated. Whaddaya think?

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -18,43 +18,50 @@ def process_response(response):
...
+def get_articles(
+    tags=None,
+    per_page=12,
+    page=1,
+    tags_exclude=[],
+    exclude=None,
+    categories=None,
+):
+    """
+    Get articles from Wordpress api
```
nottrobin: For me, the sentence describing the function is enough. I'm not too bothered about the `:param` stuff - I think the function definition + good variable names is usually clear enough for most people. But I'm also not opposed to it, so happy to leave it until such a time as the team might pick a position.

PR #16, opened 2019-03-25
Fix access to setting variables

b-m-f:
# Done
Setting were access with dotnotation. This does not working with the way they are stored on `settings.py`.


jpmartinspt:
Sounds like this is also fixed in https://github.com/canonical-webteam/blog-extension/pull/17/files.
Maybe close this one and change the version of PR 17 to `1.0.1`?
@nottrobin

PR #19, opened 2019-03-27
Improves API, better error handling, fixes url matching for Django2

b-m-f:
# DONE
- The url matching was incorrect, needed to be fixed to pass through the slug
- API was breaking, now updated
- Views had bad error handling and passed through parameters wrongly

# QA
Tests should reflect this.

### DIFF
--- a/canonicalwebteam/blog/common_view_logic.py
+++ b/canonicalwebteam/blog/common_view_logic.py
@@ -14,15 +14,9 @@ def get_index_context(page_param, articles, total_pages):
     """
 
     for article in articles:
-        try:
-            featured_image = api.get_media(article["featured_media"])
-        except Exception:
-            featured_image = None
+        featured_image = api.get_media(article["featured_media"])
 
-        try:
-            author = api.get_user(article["author"])
-        except Exception:
-            author = None
+        author = api.get_user(article["author"])
 
         category_ids = article["categories"]
 
@@ -30,10 +24,7 @@ def get_index_context(page_param, articles, total_pages):
         first_item = True
         for category_id in category_ids:
             if category_id not in category_cache:
-                try:
-                    resolved_category = api.get_category_by_id(category_id)
-                except Exception:
-                    resolved_category = None
+                resolved_category = api.get_category_by_id(category_id)
                 category_cache[category_id] = resolved_category
             if first_item:
                 article["display_category"] = category_cache[category_id]
@@ -41,10 +32,7 @@ def get_index_context(page_param, articles, total_pages):
         first_item = True
         for group_id in article["group"]:
             if group_id not in group_cache:
-                try:
-                    resolved_group = api.get_group_by_id(group_id)
-                except Exception:
-                    resolved_group = None
+                resolved_group = api.get_group_by_id(group_id)
                 group_cache[group_id] = resolved_group
             if first_item:
                 article["group"] = group_cache[group_id]
@@ -69,10 +57,7 @@ def get_article_context(article):
     :param article: Article to create context for
     """
 
-    try:
-        author = api.get_user(article["author"])
-    except Exception:
-        author = None
+    author = api.get_user(article["author"])
 
     transformed_article = logic.transform_article(
         article, author=author, optimise_images=True
@@ -80,10 +65,7 @@ def get_article_context(article):
 
     tags = article["tags"]
     tag_names = []
-    try:
-        tag_names_response = api.get_tags_by_ids(tags)
-    except Exception:
-        tag_names_response = None
+    tag_names_response = api.get_tags_by_ids(tags)
 
     if tag_names_response:
         for tag in tag_names_response:
@@ -91,12 +73,9 @@ def get_article_context(article):
 
     is_in_series = logic.is_in_series(tag_names)
 
-    try:
-        related_articles, total_pages = api.get_articles(
-            tags=tags, per_page=3, exclude=article["id"]
-        )
-    except Exception:
-        related_articles = None
+    related_articles, total_pages = api.get_articles(
+        tags=tags, per_page=3, exclude=article["id"]
+    )
 
     if related_articles:
         for related_article in related_articles:
@@ -104,10 +83,7 @@ def get_article_context(article):
 
     for group_id in article["group"]:
         if group_id not in group_cache:
-            try:
-                resolved_group = api.get_group_by_id(group_id)
-            except Exception:
-                resolved_group = None
+            resolved_group = api.get_group_by_id(group_id)
             group_cache[group_id] = resolved_group
             article["group"] = resolved_group
             break
--- a/canonicalwebteam/blog/common_view_logic.py
+++ b/canonicalwebteam/blog/common_view_logic.py
@@ -74,7 +74,7 @@ def get_article_context(article):
     is_in_series = logic.is_in_series(tag_names)
 
     related_articles, total_pages = api.get_articles(
-        tags=tags, per_page=3, exclude=article["id"]
+        tags=tags, per_page=3, exclude=[article["id"]]
     )
 
     if related_articles:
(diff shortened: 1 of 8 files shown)


nottrobin:
I've removed this from "unplanned tasks" epic and added it to "[Blog integration] Integrate module in cn.ubuntu.com" which seems more accurate. @b-m-f let me know if you disagree.

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -20,46 +26,48 @@ def get_index_context(page_param, articles, total_pages):
...
 
+        # Can these calls be bundled?
+        first_item = True
         for category_id in category_ids:
             if category_id not in category_cache:
-                category_cache[category_id] = {}
-
+                try:
+                    resolved_category = api.get_category_by_id(category_id)
+                except Exception:
```
nottrobin: It's not good practice to catch a generic `Exception`. If we must catch exceptions, we should be as explicit about it as we can be, so we're not unintentionally masking other errors. (Although I can see you simply copied this from how it used to be, I think we should take this opportunity to improve it).

Why would this category_id not exist anyway? That seems like it shouldn't happen - and therefore should raise an error?

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -20,46 +26,48 @@ def get_index_context(page_param, articles, total_pages):
...
+            if first_item:
+                article["display_category"] = category_cache[category_id]
+                first_item = False
+        first_item = True
         for group_id in article["group"]:
             if group_id not in group_cache:
-                group_cache[group_id] = {}
+                try:
+                    resolved_group = api.get_group_by_id(group_id)
+                except Exception:
```
nottrobin: Same point as for category IDs

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -20,46 +26,48 @@ def get_index_context(page_param, articles, total_pages):
 
         category_ids = article["categories"]
 
+        # Can these calls be bundled?
```
nottrobin: Maybe we should look into https://github.com/wp-graphql/wp-graphql ?

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -1,11 +1,17 @@
 from canonicalwebteam.blog import wordpress_api as api
 from canonicalwebteam.blog import logic
 
+category_cache = {}
+group_cache = {}
```
nottrobin: :heart: moving this outside the function could make a big difference.

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -20,46 +26,48 @@ def get_index_context(page_param, articles, total_pages):
...
             if category_id not in category_cache:
-                category_cache[category_id] = {}
-
+                try:
+                    resolved_category = api.get_category_by_id(category_id)
+                except Exception:
+                    resolved_category = None
+                category_cache[category_id] = resolved_category
+            if first_item:
+                article["display_category"] = category_cache[category_id]
```
nottrobin: Do you think it's neater to roll these updates into `logic.transform_article`? It seems like kinda the same concern. It could take 2 extra arguments `fetch_categories=False` and `fetch_groups=False`. What do you think?

This would also help remove code duplication in the `get_article_context` function.

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -20,46 +26,48 @@ def get_index_context(page_param, articles, total_pages):
...
         for group_id in article["group"]:
             if group_id not in group_cache:
-                group_cache[group_id] = {}
+                try:
+                    resolved_group = api.get_group_by_id(group_id)
+                except Exception:
+                    resolved_group = None
+                group_cache[group_id] = resolved_group
+            if first_item:
+                article["group"] = group_cache[group_id]
```
nottrobin: Why is the first item special?

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -49,7 +46,7 @@ def get_articles(
     return process_response(response), total_pages
 
 
-def get_article(slug, tags=[], tags_exclude=[]):
+def get_article(slug="", tags=[], tags_exclude=[]):
```
nottrobin: Why are you making the slug optional? Are you expecting to call `get_article(tags=['my-tag'])` or something, to get the first article of a tag? That would seem like a rather weird use of `get_article`.

I think it's best to leave `slug` as a mandatory argument, to signal that you're intended to get a specific article by slug.

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -65,7 +62,11 @@ def get_article(slug, tags=[], tags_exclude=[]):
 
     response = api_session.get(url)
 
-    return process_response(response)
+    data = process_response(response)
+    if len(data) > 0:
```
nottrobin: Actually, this does rather beg the question, why have `get_article` at all? If it's just going to use the same `posts` endpoint as `get_articles`, it's probably a more faithful representation of the Wordpress API to just provide the `get_articles` function, and let the client just grab the first article if that's all it wants.

Thoughts?

nottrobin:
This addresses issues in https://github.com/canonical-websites/cn.ubuntu.com/pull/303

canonicalwebteam/blog/wordpress_api.py
b-m-f: I think that this is due to the return from that endpoint. 
Making the slug not optional is key to keeping this function I feel. 
When consuming this **API** it seems nicer to have 2 separate functions, instead of overloading one with parameters.

canonicalwebteam/blog/wordpress_api.py
b-m-f: Yup! I fully agree

canonicalwebteam/blog/common_view_logic.py
b-m-f: This will be the group that gets displayed on the top of the article. It also determines the styling.
Since some articles have multiple groups, I have chosen to just pick the first one that is in the array of returned groups.

image

canonicalwebteam/blog/common_view_logic.py
b-m-f: Yup, for this use case it might word neatly. We would need the support on the wordpress site though.

canonicalwebteam/blog/common_view_logic.py
b-m-f: The main culprit that makes this necessary is the following `process_response` function in `wordpress_api`:
```
def process_response(response):
    if not response.ok:
        raise Exception("Error from api: " + response.status_code)

    return response.json()
```
Since this will throw on anything but 200 we have to catch the error somehow. A simple timeout would otherwise kill the server.

What do you think of returning `None` as the default behaviour when encountering an error?

canonicalwebteam/blog/common_view_logic.py
b-m-f: Hm, they do have slightly different logic though. Since in `get_article_context` the loop breaks immediately.
I think this might be better as a new ticket?

canonicalwebteam/blog/common_view_logic.py
nottrobin: Maybe we should just do `response.raise_for_status` and catching the native requests error?

b-m-f: Have updated it accordingly. Thanks

canonicalwebteam/blog/common_view_logic.py
nottrobin: Filed here: https://github.com/canonical-webteam/blog/issues/21

PR #22, opened 2019-04-03
Adds JSON API endpoint to fetch latest news

b-m-f:
# Done
- Adds endpoints to Flask and Django to receive latest news
- Adds new parameter to `wordpress_api` to get **sticky** posts only

# QA
Verify that the tests work. They should reflect the new URL building

### DIFF
--- a/canonicalwebteam/blog/django/views.py
+++ b/canonicalwebteam/blog/django/views.py
@@ -1,5 +1,5 @@
 from django.conf import settings
-from django.http import HttpResponseNotFound, HttpResponse
+from django.http import HttpResponseNotFound, HttpResponse, JsonResponse
 from django.shortcuts import render, redirect
 from canonicalwebteam.blog import wordpress_api as api
 from canonicalwebteam.blog import logic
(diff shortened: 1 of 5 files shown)


canonicalwebteam/blog/django/views.py
```diff
@@ -60,3 +60,29 @@ def article(request, slug):
...
+        )
+        latest_pinned_articles = api.get_articles(
+            tags=tags_id,
+            exclude=excluded_tags,
+            page=1,
+            per_page=1,
+            sticky=True,
+        )
+    except Exception as e:
+        return HttpResponse("Error: " + e, status=502)
```
jpmartinspt: If we are surfacing the error, maybe we should do it with a JSONResponse too?

canonicalwebteam/blog/flask/views.py
```diff
@@ -67,4 +67,32 @@ def article(slug):
...
+            latest_pinned_articles = api.get_articles(
+                tags=tags_id,
+                exclude=excluded_tags,
+                page=1,
+                per_page=1,
+                sticky=True,
+            )
+        except Exception as e:
+            print(e)
+            return flask.abort(502)
```
jpmartinspt: If we are surfacing the error, maybe we should do it with with jsonify too?

PR #29, opened 2019-04-16
Fix inverted parameters

b-m-f:
## Done
Function call was passing the parameters in an inverted way.

Part of https://github.com/canonical-web-and-design/jp.ubuntu.com/pull/128


tbille:
I think you can bump the version in the same PR

PR #34, opened 2019-04-25
Only display related articles that have certain tags

b-m-f:
## Done
- Only display related articles that have certain tags (extend api function with new parameter)
- Fix bug where groups where not loaded from cache

Fixes https://github.com/canonical-web-and-design/base-squad/issues/506
Fixes https://github.com/canonical-web-and-design/cn.ubuntu.com/issues/325
## QA
- Copy this new version (canonicalwebteam folder ) into an instance of https://github.com/canonical-web-and-design/cn.ubuntu.com . 
- remove the blog-module from that instances `requirements.txt`
- start the instance and go to  http://localhost:8010/blog/ubuntu-core-18-%e6%ad%a3%e5%bc%8f%e5%8f%91%e5%b8%83%ef%bc%9a%e7%94%a8%e4%ba%8e%e5%ae%89%e5%85%a8%e5%8f%af%e9%9d%a0%e7%9a%84%e7%89%a9%e8%81%94%e7%bd%91%e8%ae%be%e5%a4%87
- make sure that only chinese articles are displayed as related

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
           name: Install dependencies
           command: |
             apt install -yy python3-venv
-            pip3 install poetry
+            pip3 install poetry==0.12.11 # https://github.com/sdispater/poetry/issues/1060 can probably be unpinned when you read this.
             poetry install
       - run:
           name: Print env
--- a/canonicalwebteam/blog/common_view_logic.py
+++ b/canonicalwebteam/blog/common_view_logic.py
@@ -51,7 +51,7 @@ def get_index_context(page_param, articles, total_pages):
     }
 
 
-def get_article_context(article):
+def get_article_context(article, related_tag_ids=[]):
     """
     Build the content for the article page
     :param article: Article to create context for
@@ -73,13 +73,17 @@ def get_article_context(article):
 
     is_in_series = logic.is_in_series(tag_names)
 
-    related_articles, total_pages = api.get_articles(
+    all_related_articles, total_pages = api.get_articles(
         tags=tags, per_page=3, exclude=[article["id"]]
     )
 
-    if related_articles:
-        for related_article in related_articles:
-            related_article = logic.transform_article(related_article)
+    related_articles = []
+    if all_related_articles:
+        for related_article in all_related_articles:
+            if set(related_tag_ids).issubset(related_article["tags"]):
+                related_articles.append(
+                    logic.transform_article(related_article)
+                )
 
     for group_id in article["group"]:
         if group_id not in group_cache:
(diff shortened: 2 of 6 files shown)


canonicalwebteam/blog/django/views.py
```diff
@@ -51,13 +51,13 @@ def article_redirect(request, slug, year=None, month=None, day=None):
 
 def article(request, slug):
     try:
-        article = api.get_article(slug)
+        article = api.get_article(slug, tags_id)
```
nottrobin: Shouldn't this variable be `tag_ids`?

nottrobin: This presumably also means it should be `tags_id = settings.BLOG_CONFIG["TAGS_ID"]`, which changes the external interface and so also implies changing the settings in cn.ubuntu.com. But this should be acceptable, it just requires at least a minor version change.

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -73,20 +73,33 @@ def get_article_context(article):
...
     )
 
-    if related_articles:
-        for related_article in related_articles:
-            related_article = logic.transform_article(related_article)
+    related_articles = []
+    if all_related_articles:
+        for related_article in all_related_articles:
+            intersecting_between_related_article_tags_and_wanted = list(
+                set(related_article["tags"]) & set(related_tag_ids)
```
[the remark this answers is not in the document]
b-m-f: https://docs.python.org/3/library/stdtypes.html#frozenset.intersection

.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           name: Install dependencies
           command: |
             apt install -yy python3-venv
-            pip3 install poetry
+            pip3 install poetry==0.12.11 # https://github.com/sdispater/poetry/issues/1060 can probably be unpinned when you read this.
```
nottrobin: I read this, but I don't think it can be unpinned :stuck_out_tongue:

b-m-f: Left it for the next person to change `.circleci` :)

nottrobin:
> Copy this new version (canonicalwebteam folder ) into an instance of https://github.com/canonical-web-and-design/cn.ubuntu.com .

The way I would normally test this is by doing something like `pip install -e ../canonicalwebteam.blog`, or adding `-e ../canonicalwebteam.blog` or `git+git@github.com:org/canonicalwebteam.blog#egg=canonicalwebteam.blog` into `requirements.txt`.

All of these workflows are impossible without a `setup.py`.

I suspect, once we've moved all our projects over to Poetry then this problem will go away (we'll no longer be doing `pip install` anywhere), but until then, I think we should probably preserve `setup.py` so as to enable these workflows. We'll just have to make sure we do version updates etc. in both places.

@b-m-f do you agree? I'm creating an issue, where we can continue discussion.

PR #36, opened 2019-04-30
Featured posts on first page

b-m-f:
# Done
- Fetches 3 featured posts on the first page of the blog.

# QA
Run the tests with `poetry install && poetry run pytest`

### DIFF
--- a/canonicalwebteam/blog/common_view_logic.py
+++ b/canonicalwebteam/blog/common_view_logic.py
@@ -5,49 +5,67 @@ category_cache = {}
 group_cache = {}
 
 
-def get_index_context(page_param, articles, total_pages):
+def transform_index_article(article):
+    """
+    This transforms any given article from the wordpress API
+    into an object that includes all information for the templates,
+    some of which will be fetched from the Wordpress API
+    """
+    featured_image = api.get_media(article["featured_media"])
+
+    author = api.get_user(article["author"])
+
+    category_ids = article["categories"]
+
+    # Can these calls be bundled?
+    first_item = True
+    for category_id in category_ids:
+        if category_id not in category_cache:
+            resolved_category = api.get_category_by_id(category_id)
+            category_cache[category_id] = resolved_category
+        if first_item:
+            article["display_category"] = category_cache[category_id]
+            first_item = False
+
+    first_item = True
+    for group_id in article["group"]:
+        if group_id not in group_cache:
+            resolved_group = api.get_group_by_id(group_id)
+            group_cache[group_id] = resolved_group
+        if first_item:
+            article["group"] = group_cache[group_id]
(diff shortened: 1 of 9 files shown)


canonicalwebteam/blog/django/views.py
```diff
@@ -15,16 +15,39 @@
...
+            articles, total_pages = api.get_articles(
+                tags=tag_ids, page=page_param
+            )
+            featured_articles = []
+
     except Exception as e:
         return HttpResponse("Error: " + e, status=502)
 
-    context = get_index_context(page_param, articles, total_pages)
+    print(get_index_context)
```
nottrobin: Did you accidentally leave this print statement in?

canonicalwebteam/blog/django/views.py
```diff
@@ -15,16 +15,39 @@
...
     try:
-        articles, total_pages = api.get_articles(
-            tags=tag_ids, tags_exclude=excluded_tags, page=page_param
-        )
+        if page_param == "1":
+            featured_articles, total_pages = api.get_articles(
+                tags=tag_ids,
+                tags_exclude=excluded_tags,
+                page=page_param,
+                sticky="true",
```
nottrobin: I assume this is what makes them "featured"?

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -5,15 +5,15 @@
...
+def get_index_context(page_param, articles, total_pages, featured_articles=[]):
     """
     Build the content for the index page
     :param page_param: String or int for index of the page to get
     :param articles: Array of articles
     :param articles: String of int of total amount of pages
     """
 
-    for article in articles:
+    def transform_index_article(article):
```
nottrobin: Why  is this nested inside `get_index_context`? I don't think it needs to be does it? It's only needs the article that's passed to it right?

If you've no objection, could you please move this function to being top level? (And ideally give it a docstring)

canonicalwebteam/blog/django/views.py
b-m-f: oh, thanks :)

canonicalwebteam/blog/django/views.py
b-m-f: Yes, exactly.

nottrobin:
I get:

``` bash
$ poetry run pytest
Creating virtualenv canonicalwebteam.blog-py3.6 in /home/robin/.cache/pypoetry/virtualenvs
                                                           
[FileNotFoundError]                         
[Errno 2] No such file or directory: 'pytest': 'pytest'  
                                                           
run <args> (<args>)...
```

Is that expected? Is that what you meant when you said you were giving up on tests?

PR #40, opened 2019-05-02
Group pages

b-m-f:
# Done
- adds new view function for group pages
- routes to a predefined set of groups
- adds category parameter to group view
- Looks up group templates by slug

# Issues
Fixes https://github.com/canonical-web-and-design/base-squad/issues/455

# QA
`poetry install && poetry run pytest`


canonicalwebteam/blog/django/urls.py
```diff
@@ -49,6 +50,22 @@ def to_url(self, value):
     path(r"/<yyyy:year>/<str:slug>", article_redirect),
     path(r"/feed", feed),
     path(r"/latest-news", latest_news),
+    path(
+        r"/cloud-and-server", group, {"slug": "cloud-and-server"}, name="group"
```
nottrobin: I feel rather like defining these specific URLs should be beyond the scope of this module, as it seems like a specific use-case. It feels wrong for this module to know anything about e.g. the "cloud-and-server" group specifically.

I think this module should provide the `group` view but that these URLs should be kept in www.ubuntu.com's `urls.py`.

If this module were to provide generic group pages, I would have thought they'd take the form `/groups/{name}`, but we probably shouldn't create that endpoint yet as we don't have a use-case for it.

README.md
```diff
@@ -45,10 +44,17 @@ BLOG_CONFIG = {
     "TAG_NAME": "TAG NAME FOR GENERATING A FEED",
 }
 ```
-- You can now use the data from the blog. To display it the module expects templates at `blog/index.html`, `blog/article.html` and `blog/blog-card.html`. Inspiration can be found at https://github.com/canonical-websites/jp.ubuntu.com/tree/master/templates/blog.
-
 - Run your project and verify that the blog is displaying at the path you specified (f.e. '/blog')
 
+## Templates
+
+- You can now use the data from the blog. To display it the module expects templates at `blog/index.html`, `blog/article.html`, `blog/blog-card.html` . Inspiration can be found at https://github.com/canonical-websites/jp.ubuntu.com/tree/master/templates/blog.
```
nottrobin: As mentioned below, I think we should move the URL definitions to www.ubuntu.com, and so this block should probably instead describe how to implement the `group` view.

b-m-f: I have updated the module now. It takes a slug and template path and can be configured by the consuming app.

PR #42, opened 2019-05-10
Adds news context generator function for group pages

b-m-f:
The main group on articles was incorrect on group pages.
The initial plan was to fix this on the frontend.
After some reflection on the design, it became clear that implementing this on the backend is the cleaner solution

## QA
- Use https://github.com/canonical-web-and-design/www.ubuntu.com and replace `canonicalwebteam.blog==1.5.1` with `git+https://github.com/b-m-f/blog-extension@add-correct-group-to-articles-for-group-pages#egg=canonicalwebteam.blog` in `requirements.txt`
- `./run` the site
- head to `localhost;8001/blog/internet-of-things` and verify that all articles are displayed with that group on the article card

### DIFF
--- a/canonicalwebteam/blog/common_view_logic.py
+++ b/canonicalwebteam/blog/common_view_logic.py
@@ -5,10 +5,10 @@ category_cache = {}
 group_cache = {}
 
 
-def transform_index_article(article):
+def get_complete_article(article, group=None):
     """
-    This transforms any given article from the wordpress API
-    into an object that includes all information for the templates,
+    This returns any given article from the wordpress API
+    as an object that includes all information for the templates,
     some of which will be fetched from the Wordpress API
     """
     featured_image = api.get_media(article["featured_media"])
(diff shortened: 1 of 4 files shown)


canonicalwebteam/blog/common_view_logic.py
```diff
@@ -27,14 +27,20 @@ def transform_index_article(article):
...
 
-    first_item = True
-    for group_id in article["group"]:
-        if group_id not in group_cache:
-            resolved_group = api.get_group_by_id(group_id)
-            group_cache[group_id] = resolved_group
-        if first_item:
-            article["group"] = group_cache[group_id]
-            first_item = False
+    if group_slug is not None:
```
jpmartinspt: Can't this just be `if group_slug:`?

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -5,10 +5,10 @@
 group_cache = {}
 
 
-def transform_index_article(article):
+def get_complete_article(article, group_slug=None):
```
jpmartinspt: The arguments name is `group_slug` but you are effectively passing the group id when you call the function. Might be worth renaming so it doesn't get confusing.

PR #43, opened 2019-05-10
Add archives

b-m-f:
Depends on https://github.com/canonical-web-and-design/canonicalwebteam.blog/pull/42

Adds archives endpoint

## QA
- Use https://github.com/canonical-web-and-design/www.ubuntu.com and replace `canonicalwebteam.blog==1.5.1` with `git+https://github.com/b-m-f/blog-extension@add-archives#egg=canonicalwebteam.blog` in `requirements.txt`
- `./run` the site
- head to `localhost;8001/blog/archives?year=2019&month=1` and verify that all articles displayed are from that month. Can be verified by checking against https://blog.ubuntu.com/archives?year=2019&month=1

### DIFF
--- a/canonicalwebteam/blog/django/views.py
+++ b/canonicalwebteam/blog/django/views.py
@@ -1,6 +1,6 @@
-from django.conf import settings
-from django.http import HttpResponseNotFound, HttpResponse, JsonResponse
-from django.shortcuts import render, redirect
+from datetime import datetime
+
+
 from canonicalwebteam.blog import wordpress_api as api
 from canonicalwebteam.blog import logic
 from canonicalwebteam.blog.common_view_logic import (
@@ -8,6 +8,10 @@ from canonicalwebteam.blog.common_view_logic import (
     get_article_context,
     get_group_page_context,
 )
+from dateutil.relativedelta import relativedelta
+from django.conf import settings
+from django.http import HttpResponseNotFound, HttpResponse, JsonResponse
+from django.shortcuts import render, redirect
 
 tag_ids = settings.BLOG_CONFIG["TAG_IDS"]
 excluded_tags = settings.BLOG_CONFIG["EXCLUDED_TAGS"]
@@ -84,6 +88,62 @@ def group(request, slug, template_path):
     context["category"] = {"slug": category_param}
 
     return render(request, template_path, context)
+
+
+def archives(request, template_path="blog/archives.html"):
+    try:
+        page = request.GET.get("page", default="1")
+        category = request.GET.get("category", default="")
+        group = request.GET.get("group", default="")
+        month = request.GET.get("month", default="")
+        year = request.GET.get("year", default="")
+
+        groups = []
+        categories = []
+
+        if group:
+            group = api.get_group_by_slug(group)
+            groups.append(group["id"])
+
+        if category:
+            category = api.get_category_by_slug(category)
+            categories.append(category["id"])
+
+        after = ""
+        before = ""
+        if year:
+            year = int(year)
+            if month:
+                month = int(month)
+                after = datetime(year=year, month=month, day=1)
+                before = after + relativedelta(months=1)
+            else:
+                after = datetime(year=year, month=1, day=1)
+                before = datetime(year=year, month=12, day=31)
+
+        articles, total_pages = api.get_articles(
+            tags=tag_ids,
+            tags_exclude=excluded_tags,
+            page=page,
+            groups=groups,
+            categories=categories,
+            after=after,
+            before=before,
+        )
+
+        if group:
+            context = get_group_page_context(
+                page, articles, total_pages, group
+            )
+        else:
+            context = get_index_context(page, articles, total_pages)
+
+        context["title"] = blog_title
+
+        return render(request, template_path, context)
+
+    except Exception as e:
+        return HttpResponse("Error: " + e, status=502)
 
 
 def feed(request):
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,4 @@
+1.5.4: Adds archives Endpoint for Django
 1.5.3: Adds current category object to the page context for filtering
 1.5.2: Get correct article group on group page
 1.5.1: Ability to load articles for specific groups only, filtering them by category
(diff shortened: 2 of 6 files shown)


canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,68 @@ def group(request, slug, template_path):
     return render(request, template_path, context)
 
 
+def archives(request, template_path="blog/archives.html"):
+    try:
+        page_param = request.GET.get("page", default="1")
```
jpmartinspt: Can we just remove the `_param` suffix and reassign if needed?

b-m-f: sure :+1:

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,68 @@ def group(request, slug, template_path):
...
+        category_param = request.GET.get("category", default="")
+        group_param = request.GET.get("group", default="")
+        month_param = request.GET.get("month", default="")
+        year_param = request.GET.get("year", default="")
+
+        group = None
+        group_id = ""
+        if group_param != "":
+            group = api.get_group_by_slug(group_param)
+            group_id = group["id"]
```
jpmartinspt: I would declare `groups and categories = []` beforehand and then append the id to it inside the if.

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,68 @@ def group(request, slug, template_path):
...
+
+        category_id = ""
+        if category_param != "":
+            category = api.get_category_by_slug(category_param)
+            category_id = category["id"]
```
jpmartinspt: I would declare groups and categories = [] beforehand and then append the id to it inside the if.

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,68 @@ def group(request, slug, template_path):
...
+                before = after + relativedelta(months=1)
+            else:
+                after = datetime(year=year, month=1, day=1)
+                before = after + relativedelta(years=1)
+
+        articles, total_pages = api.get_articles(
+            tags=tag_ids,
+            tags_exclude=excluded_tags,
+            page=page_param,
+            groups=[group_id],
```
jpmartinspt: This line would be `groups=groups,` in case you swap to use lists.

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,68 @@ def group(request, slug, template_path):
...
+            categories=[category_id],
```
jpmartinspt: This line would be `categories=categories,` in case you swap to use lists.

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,68 @@ def group(request, slug, template_path):
     return render(request, template_path, context)
 
 
+def archives(request, template_path="blog/archives.html"):
+    try:
+        page_param = request.GET.get("page", default="1")
+        category_param = request.GET.get("category", default="")
+        group_param = request.GET.get("group", default="")
+        month_param = request.GET.get("month", default="")
```
jpmartinspt: For `month` and `year` you could probably just try to `int` cast here since we are broadly handling the exceptions?

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,68 @@ def group(request, slug, template_path):
...
+            groups=[group_id],
+            categories=[category_id],
+            after=after,
+            before=before,
+        )
+
+    except Exception as e:
+        return HttpResponse("Error: " + e, status=502)
+
+    if group:
```
jpmartinspt: Can we move 143-152 inside the try block?

canonicalwebteam/blog/django/views.py
b-m-f: I have added sth like this (a memory based cache) in https://github.com/canonical-web-and-design/canonicalwebteam.blog/pull/45 . 

Mind giving that review? Then I can rebase this branch on top of it.

canonicalwebteam/blog/django/views.py
b-m-f: This might break when year or month are `""`(empty string.)

canonicalwebteam/blog/django/views.py
b-m-f: Works :)

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,65 @@ def group(request, slug, template_path):
...
 
+def archives(request, template_path="blog/archives.html"):
+    try:
+        page = request.GET.get("page", default="1")
+        category = request.GET.get("category", default="")
+        group = request.GET.get("group", default="")
+        month = request.GET.get("month", default="")
+        year = request.GET.get("year", default="")
+
+        groups = categories = []
```
jpmartinspt: You need to split this into two lines otherwise `groups` and `categories` are pointing at the same list.

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,63 @@ def group(request, slug, template_path):
...
+
+        if year:
+            year = int(year)
+            if month:
+                month = int(month)
+            after = datetime(year=year, month=1, day=1)
+            before = datetime(year=year, month=12, day=31)
+            if month != "":
+                after = datetime(year=year, month=month, day=1)
+                before = after + relativedelta(months=1)
```
tbille:
```python
            if month:
                month = int(month)
                after = datetime(year=year, month=month, day=1)
                before = after + relativedelta(months=1)
            else:
                after = datetime(year=year, month=1, day=1)
                before = datetime(year=year, month=12, day=31)
```

canonicalwebteam/blog/django/views.py
```diff
@@ -86,6 +90,63 @@ def group(request, slug, template_path):
     return render(request, template_path, context)
 
 
+def archives(request, template_path="blog/archives.html"):
+    try:
```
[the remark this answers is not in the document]
tbille: After discussion with @jpmartinspt . I still think it's not a right solution. But for now we can keep it this way.

jpmartinspt:
If you don't pass any year and month it crashes, looking at current blog it should just display all posts.
Setting after and before in the view to empty/none values should fix this?

PR #45, opened 2019-05-13
Adds upcoming endpoint and adds memory based caching for categories a…

b-m-f:
## Done
- adds `/upcoming` endpoint
- adds memory based caching for categories and groups

Fixes https://github.com/canonical-web-and-design/base-squad/issues/524

## QA
- Replace `canonicalwebteam.blog` with `https://github.com/b-m-f/canonicalwebteam.blog@add-upcomming-endpoint#egg=canonicalwebteam.blog` in your `requirements.txt`
- `./run`
- include `{{articles}}` in `blog/upcoming.html` template- go to https://localhost:8001/blog/upcoming and make sure that articles that are listed are the same as on https://blog.ubuntu.com/upcoming

### DIFF
--- a/canonicalwebteam/blog/django/views.py
+++ b/canonicalwebteam/blog/django/views.py
@@ -105,11 +105,33 @@ def topic(request, slug, template_path):
 
     except Exception as e:
         return HttpResponse("Error: " + e, status=502)
-
     context = get_topic_page_context(page_param, articles, total_pages)
     context["title"] = blog_title
 
     return render(request, template_path, context)
+
+
+def upcoming(request):
+    try:
+        page_param = request.GET.get("page", default="1")
+
+        events = api.get_category_by_slug("events")
+        webinars = api.get_category_by_slug("webinars")
+
+        articles, total_pages = api.get_articles(
+            tags=tag_ids,
+            tags_exclude=excluded_tags,
+            page=page_param,
+            categories=[events["id"], webinars["id"]],
+        )
+
+    except Exception as e:
+        return HttpResponse("Error: " + e, status=502)
+
+    context = get_index_context(page_param, articles, total_pages)
+    context["title"] = blog_title
+
+    return render(request, "blog/upcoming.html", context)
 
 
 def archives(request, template_path="blog/archives.html"):
--- a/canonicalwebteam/blog/wordpress_api.py
+++ b/canonicalwebteam/blog/wordpress_api.py
@@ -121,7 +121,8 @@ def get_group_by_slug(slug):
     if not response.ok:
         return None
     try:
-        return process_response(response)[0]
+        groups_response = process_response(response)
+        return groups_response[0]
     except Exception:
         return None
 
@@ -130,8 +131,8 @@ def get_group_by_id(id):
     url = "".join([API_URL, "/group/", str(id)])
 
     response = api_session.get(url)
-
-    return process_response(response)
+    group = process_response(response)
+    return group
 
 
 def get_category_by_slug(slug):
@@ -142,7 +143,8 @@ def get_category_by_slug(slug):
     if not response.ok:
         return None
     try:
-        return process_response(response)[0]
+        category = process_response(response)
+        return category[0]
     except Exception:
         return None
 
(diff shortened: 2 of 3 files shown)


canonicalwebteam/blog/django/views.py
```diff
@@ -105,13 +105,35 @@ def topic(request, slug, template_path):
...
 
     return render(request, template_path, context)
 
 
+def upcoming(request):
+    try:
+        page_param = request.GET.get("page", default="1")
+
+        events_id = api.get_category_by_slug("events")["id"]
+        print(events_id)
```
tbille: remove print

canonicalwebteam/blog/django/views.py
```diff
@@ -105,13 +105,35 @@ def topic(request, slug, template_path):
...
     context["title"] = blog_title
 
     return render(request, template_path, context)
 
 
+def upcoming(request):
+    try:
+        page_param = request.GET.get("page", default="1")
+
+        events_id = api.get_category_by_slug("events")["id"]
```
tbille: `events = api.get_category_by_slug("events")`

canonicalwebteam/blog/django/views.py
```diff
@@ -105,13 +105,35 @@ def topic(request, slug, template_path):
...
+        print(events_id)
+        webinars_id = api.get_category_by_slug("webinars")["id"]
```
tbille: `webinars = api.get_category_by_slug("webinars")`

canonicalwebteam/blog/django/views.py
```diff
@@ -105,13 +105,35 @@ def topic(request, slug, template_path):
...
+
+        articles, total_pages = api.get_articles(
+            tags=tag_ids,
+            tags_exclude=excluded_tags,
+            page=page_param,
+            categories=[events_id, webinars_id],
```
tbille: `categories=[events["id"], webinars["id"]],`

canonicalwebteam/blog/django/views.py
```diff
@@ -105,13 +105,35 @@ def topic(request, slug, template_path):
...
+        )
+
+    except Exception as e:
+        return HttpResponse("Error: " + e, status=502)
+    context = get_index_context(page_param, articles, total_pages)
```
tbille: add line before, let's make the code breath a little :smile:

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -13,6 +13,10 @@
 tags = {}
 
 
+categories = {}
+groups = {}
```
tbille: is this a cache?

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -114,45 +118,63 @@ def get_categories():
...
     url = "".join([API_URL, "/group?", f"slug={slug}"])
 
     response = api_session.get(url)
 
     if not response.ok:
         return None
     try:
-        return process_response(response)[0]
+        group = process_response(response)[0]
+        groups[slug] = group
```
tbille:
```
        group = process_response(response)
        groups[slug] = group[0]
```

canonicalwebteam/blog/wordpress_api.py
b-m-f: Yup, preliminary one.

tbille: we already cache the api calls. Let's not have a waterfall of caching this will lead us to issues for refreshing (etc.). The applications should be stateless. If we want to implement one day some memory caching we can think about it and design it.

tbille: I understand why you are adding this, but I think by integrating the changes that we have done on s.io/blog we shouldn't need all this caching

b-m-f: Yup, I had the same argument. Joao mentioned though that we could build this out into a smarter cache.

Blog.ubuntu has a similiar strategy to cache the categories and groups both by ID and slug, to make looksup faster. Since we will do this lookup quite often, I think it could be worth it.

The only problem here is cache invalidation, which would need to occor and was the only concern.
But we disregarded this, as it is unlikely for now that this will change in Wordpress.

Anything I forgot @jpmartinspt ?

jpmartinspt: I think it's fine, but per function call instead of in the global namespace. Might be useful in functions where we fetch multiple posts, but if we update the API calls to include related data it won't be of much use anymore.

b-m-f: Have just removed it now :) Lets go ahead and implement the embedded api.

PR #46, opened 2019-05-14
Adds topic pages

b-m-f:
Adds topic endpoint

Fixes https://github.com/canonical-web-and-design/base-squad/issues/526

## QA
- Use https://github.com/canonical-web-and-design/www.ubuntu.com and replace `canonicalwebteam.blog==1.5.X` with `git+https://github.com/b-m-f/blog-extension@add-topic-pages#egg=canonicalwebteam.blog` in `requirements.txt`
- `./run` the site
- create a template at `blog/snappy.html` or copy the existing `blog/index.html` to that file
- add the following code to `urls.py`
```
    path(
        r"blog/topics/snappy",
        topic,
        {"slug": "snappy", "template_path": "blog/snappy.html"},
        name="topic",
    ),

```

- head to `localhost;8001/blog/topics/snappy` and verify that all articles displayed are from that month. Can be verified by checking against https://blog.ubuntu.com/topics/snappy

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,19 @@ urlpatterns = [
     )
 ```
 
+#### Topic pages
+- Topic pages are optional as well and can be enabled by using the view `canonicalwebteam.blog.django.views.topic`. The view takes the topic slug to fetch data for and a template path to load the correct template from.
+
+**urls.py**
+```python
+path(
+		r"blog/topic/kubernetes",
+		topic,
+		{"slug": "kubernetes", "template_path": "blog/kubernetes.html"},
+		name="topic",
+),
+```
+
 ## Templates
 
 - You can now use the data from the blog. To display it the module expects templates at `blog/index.html`, `blog/article.html`, `blog/blog-card.html` . Inspiration can be found at https://github.com/canonical-websites/jp.ubuntu.com/tree/master/templates/blog.
--- a/canonicalwebteam/blog/django/urls.py
+++ b/canonicalwebteam/blog/django/urls.py
@@ -48,9 +48,9 @@ urlpatterns = [
     path(r"/<yyyy:year>/<mm:month>/<dd:day>/<str:slug>", article_redirect),
     path(r"/<yyyy:year>/<mm:month>/<str:slug>", article_redirect),
     path(r"/<yyyy:year>/<str:slug>", article_redirect),
+    path(r"/<str:slug>", article, name="article"),
+    path(r"/archives", archives),
     path(r"/feed", feed),
     path(r"/latest-news", latest_news),
-    path(r"/archives", archives),
-    path(r"/<str:slug>", article, name="article"),
     path(r"", index),
 ]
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -63,6 +63,19 @@ urlpatterns = [
     )
 ```
 
+#### Topic pages
+- Topic pages are optional as well and can be enabled by using the view `canonicalwebteam.blog.django.views.topic`. The view takes the topic slug to fetch data for and a template path to load the correct template from.
```
nottrobin: I'm not sure we should do topics this way. Unlike with Categories, where it was top-level URLs that couldn't be provided generically (`blog/cloud-and-server`), these URLs are easily generalisable (`blog/topics/{topic_name}`).

So in this case I think maybe we should just make this a setting that you add to `webapp/settings.py`, like `ENABLE_TOPIC_PAGES: True`. What do you think?

b-m-f: Sounds good. The template will then correspond to the slug. Makes it generic enough

nottrobin: huh? you need a different template per topic page? Maybe that's why you did it this way?

nottrobin: In that case, maybe your solution is better

b-m-f: Updated it :)

canonicalwebteam/blog/django/urls.py
```diff
@@ -48,9 +49,10 @@ def to_url(self, value):
     path(r"/<yyyy:year>/<mm:month>/<dd:day>/<str:slug>", article_redirect),
     path(r"/<yyyy:year>/<mm:month>/<str:slug>", article_redirect),
     path(r"/<yyyy:year>/<str:slug>", article_redirect),
+    path(r"/<str:slug>", article, name="article"),
```
nottrobin: If this is up here, won't it clobber `/feed` and `/latest-news`?

README.md
nottrobin: Thanks, sorry

canonicalwebteam/blog/django/urls.py
b-m-f: It does indeed :) Fixed

README.md
```diff
@@ -63,6 +63,11 @@ urlpatterns = [
     )
 ```
 
+#### Topic pages
+- Topic pages are optional as well and can be enabled by setting `ENABLE_TOPIC_PAGES` to `true` in the `BLOG_CONFIG` described above.
+  The template for a topic will be loaded by its slug. E.g. `http://localhost:8001/blog/topic/snappy` will load the temlate at `templates/blog/snappy.html`
```
nottrobin: I thought you said you'd switched this back?

README.md
```diff
@@ -63,6 +63,19 @@ urlpatterns = [
     )
 ```
 
+#### Topic pages
+- Topic pages are optional as well and can be enabled by using the view `canonicalwebteam.blog.django.views.topic`. The view takes the topic slug to fetch data for and a template path to load the correct template from.
+
+**urls.py**
+```python
+path(
+		r"blog/topic/kubernetes",
```
nottrobin: Oh just one tiny point - could you make this `/topics`?

b-m-f: will do

PR #53, opened 2019-05-17
Adds the next 3 upcoming events and webinars to the index page

b-m-f:
# Done
- Adds the next 3 upcoming events and webinars to the index page

Fixes https://github.com/canonical-web-and-design/base-squad/issues/520

## QA
- Use https://github.com/canonical-web-and-design/www.ubuntu.com and replace `canonicalwebteam.blog==1.5.X` with `git+https://github.com/b-m-f/blog-extension@add-upcoming-to-index#egg=canonicalwebteam.blog` in `requirements.txt`
- `./run` the site
- add `{{upcoming}}` to `blog/index.html` and make sure that the data is present when going to `/blog`. It should be 3 entries

### DIFF
--- a/canonicalwebteam/blog/django/views.py
+++ b/canonicalwebteam/blog/django/views.py
@@ -20,9 +20,10 @@ blog_title = settings.BLOG_CONFIG["BLOG_TITLE"]
 tag_name = settings.BLOG_CONFIG["TAG_NAME"]
 
 
-def index(request):
+def index(request, enable_upcoming=True):
     page_param = request.GET.get("page", default="1")
     category_param = request.GET.get("category", default="")
+    upcoming = []
 
     try:
         category_id = ""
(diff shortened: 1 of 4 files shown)


jpmartinspt:
What about the flask view? It seems to be drifting apart from the django view.
Also it's probably worth having a way of activating the upcoming feature so the request is not made when it's not needed?

b-m-f:
For the Flask views I added https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/61.

I have added a parameter to disable the upcoming posts though.

canonicalwebteam/blog/django/views.py
```diff
@@ -47,6 +47,19 @@ def index(request):
                 page=page_param,
                 categories=[category_id],
             )
+            if upcoming:
+                events = api.get_category_by_slug("events")
+                webinars = api.get_category_by_slug("webinars")
+                upcoming, _ = api.get_articles(
```
jpmartinspt: We should name this differently from the function argument, or rename the argument.

canonicalwebteam/blog/django/views.py
```diff
@@ -47,6 +47,19 @@ def index(request):
...
+                    tags=tag_ids,
+                    tags_exclude=excluded_tags,
+                    page=page_param,
+                    per_page=3,
+                    categories=[events["id"], webinars["id"]],
+                )
+            else:
```
jpmartinspt: Could be declared up top as an empty list to avoid the else.

PR #55, opened 2019-05-20
Improve testing

b-m-f:
# Done
Improves the test coverage for the blog modules wordpress api wrapper
This is done by testing each function individually and using [vcrpy](https://github.com/kevin1024/vcrpy) to record actual HTTP requests and their results as fixtures for the tests.

Fixes https://github.com/canonical-web-and-design/base-squad/issues/535

**When reviewing, feel free to collapse all the fixtures. They are autogenerated and map to the wordpress API responses.**

# QA
Run `poetry run pytest`


tests/test_wordpress_api.py
```diff
@@ -116,157 +30,162 @@ def test_getting_all_articles(self, get):
...
-            + "per_page=12"
-            + "&tags="
-            + "&page=1"
-            + "&group="
-            + "&tags_exclude=1234,5678"
-            + "&categories="
-            + "&exclude="
-            + "&author="
+        url = api.build_get_articles_url(page=3, per_page=24)
+        self.assertEqual(
```
nottrobin: I think this test is getting a little too specific - checking the exact format of the built URL certainly feels like testing internals.

Personally I'd suggest you treat `build_get_articles_url` as an internal function / private method of `api` (maybe rename it to `_build_get_articles_url` [to signify this](https://docs.python.org/3/tutorial/classes.html#private-variables)) and so say that it shouldn't be tested at all. Instead, I think it would be ideal to only test `get_articles` and `get_articles_with_metadata` - as these form the important external interface of the API library - and just make sure they return the expected articles. Of course this means you have to have fixtures for all the different filtered API responses.

tests/test_wordpress_api.py
```diff
@@ -116,157 +30,162 @@ def test_getting_all_articles(self, get):
...
+        self.assertEqual(tag["name"], name)
 
     @patch("canonicalwebteam.http.CachedSession.get")
     def test_getting_articles_fail(self, get):
-
         get.return_value = FailingMockResponse()
 
         with self.assertRaises(Exception):
-            api.get_articles()
+            api.get.articles()
```
nottrobin: What's this doing?

pyproject.toml
```diff
@@ -21,6 +21,7 @@ flask_testing = "^0.7.1"
 pytest = "^4.4"
 django = "^2.2"
 Flask = "^1.0.1"
+vcrpy = "^2.0"
```
nottrobin: `vcrpy` looks interesting. Did you consider writing and committing a script to regenerate the fixtures? I think that would be valuable - I assume this is the sort of thing that `vcr` makes super easy.

tests/test_wordpress_api.py
b-m-f: Good idea! I think its worth the effort, as this is such a core functionality

pyproject.toml
b-m-f: Yup, basically `rm -rf fixtures && poetry run pytest` to regenerate everything

tests/test_wordpress_api.py
b-m-f: So, I have left this unittest in for now, as the function itself needs updating. 
No harm in having an extra unit-test, no?

tests/test_wordpress_api.py
b-m-f: Removed, probably a typo

pyproject.toml
b-m-f: Ill add it to the readme instead :)

tests/test_wordpress_api.py
nottrobin: There is harm. The harm is that if some well-meaning soul wants to refactor the code in a way that doesn't effect the actual external interface of the module, their job will be that much harder. They have to either feel empowered to delete this test, or attempt to rewrite it.

But even if you were going to leave it in place, I would say it's better to do a test something like:

``` python3
self.assertTrue('page=3' in urlparse(url).query)
```

rather than testing the exact format of the URL output. Then at least you're accommodating changes in the order of the parameters, will have zero effect on the functionality.

b-m-f: While I agree in general, I think this function has a lot of responsibility and should be guarded.

I do like the `assertTrue` logic a lot better though, it makes it clearer what the test does. I will update it now :)

nottrobin: I still want to defend the principle that in a well-written module, only the external interface should need to be tested. This is because a neat module won't have gold-plated functions, each function will only do as much as is needed by the actual functions of the module. Therefore, it should be possible to test every line of code in the module by simply testing each of the module's external functions. If this isn't possible, code should be removed from the internals.

In very complex modules, I can see that there might be cases where trying to test an internal line of code from the outside might require a very convoluted test, which might then be very hard to maintain. In this case there could be an argument for testing an internal function so as to keep tests simpler and maintainable.

But in this case I don't think this argument holds. Everything you're testing with `build_get_articles_url` could trivially be instead tested with `get_articles` or `get_articles_with_metadata` - and in fact they really should be. The really important thing to test here is that `get_articles(page=3)` works, because that is the actual interface that clients will use, which you're not actually testing.

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -83,7 +83,7 @@ def get_articles_with_metadata(**kwargs):
 
     :returns: response, metadata dictionary
     """
-    url = build_get_article_url(**kwargs)
+    url = build_get_articles_url(**kwargs)
```
nottrobin: Related to what we are discussion about testing, I think you should switch around these interfaces. `get_articles` and `get_articles_with_metadata` should take *explicit* arguments for the functions they expect to fulfill - `page` etc.

Then, I think it would be reasonable to change `build_get_articles_url` to be flexible to its parameters, as it doesn't actually need any specific knowledge - all it does is:

- Convert arrays into comma-separated strings
- Convert booleans into `name={bool}` strings
- Compile the query string

So I actually think you could happily genericise `build_get_articles_url` into a `_build_query` function, which takes a `parameters` argument (I consider `**kwargs` to be a bit of a potentially-confusing hack that we should avoid if we can), and then does the transformations I describe above.

Again, I don't think this function itself should need testing, its usage by `get_articles` and every other extenal function that makes an API call with a query string (whose *entire* interface should be tested) should provide quite enough of a test.

tests/test_wordpress_api.py
b-m-f: Click! Got it now and updated the tests

canonicalwebteam/blog/wordpress_api.py
b-m-f: Yup, @jpmartinspt mentioned this before. I think Ill ask him for some help on the urllib and reopen that issue

b-m-f: @nottrobin this is the issue https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/15 .

Id rather not include this into this PR as well :)

tests/test_common_view_logic.py
```diff
@@ -1,389 +1,182 @@
...
-            },
-            {
-                "featured_media": "test2",
-                "author": "test2",
-                "categories": [2, 3],
-                "group": [1],
-                "tags": ["test2"],
-            },
+    @vcr.use_cassette("fixtures/vcr_cassettes/get_complete_article.yaml")
+    def test_get_complete_article(self):
```
tbille: I feel here you are just testing that your fixtures are well written/

If you want to have exhaustive you should try to break your application instead:

- What happens if you give an empty object?
- What happens if there is no catogory/group/author?

Each endpoint IMO should be in it's own class and file, even if in the meantime there will be only one function per class. Otherwise you might end with a very very long test suite file.

tests/test_common_view_logic.py
```diff
@@ -1,389 +1,182 @@
...
+    get_complete_article,
     get_index_context,
+    get_group_page_context,
+    get_topic_page_context,
     get_article_context,
 )
+from canonicalwebteam.http import CachedSession
+
+# make sure that timeouts do not hinder tests
+api.api_session = CachedSession(fallback_cache_duration=3600, timeout=(1, 100))
```
tbille: Remove cache for tests

`api.api_session = Session()`

tests/test_common_view_logic.py
```diff
@@ -1,389 +1,182 @@
...
-                "tags": ["test2"],
-            },
+
+        featured, _ = api.get_articles(sticky=True)
+
+        # TODO: TDD refactor of function signature, to get
+        # the whole index without any params
+        index_context = get_index_context(1, articles, "2", featured)
+
+        # page information is correct
```
tbille: please no comments. Same for the rest

tests/test_common_view_logic.py
```diff
@@ -1,389 +1,182 @@
...
-                    "categories": [4, 5],
-                    "group": [4],
-                    "tags": ["test_related_article"],
-                }
-            ],
-            2,
+
+        featured, _ = api.get_articles(sticky=True)
+
+        # TODO: TDD refactor of function signature, to get
```
tbille: When you add TODOs please open and issue and link it in the code

tests/test_wordpress_api.py
```diff
@@ -1,20 +1,13 @@
...
-
-    def json(self):
-        return ["hello_test"]
-
-    def raise_for_status(*args):
-        pass
-
-    headers = {"X-WP-TotalPages": 12}
+# make sure that timeouts do not hinder tests
+api.api_session = CachedSession(fallback_cache_duration=3600, timeout=(1, 100))
```
tbille: Same comment. Replace with Session instead of using cache

tests/test_wordpress_api.py
```diff
@@ -27,245 +20,181 @@ def raise_for_status(*args):
...
-        article = api.get_article(
-            slug="test", tags=[1234, 5678], tags_exclude=[9876]
+    @vcr.use_cassette("fixtures/vcr_cassettes/articles_with_metadata.yaml")
+    def test_get_articles_with_metadata(self):
+        articles, metadata = api.get_articles_with_metadata()
+        self.assertIsNotNone(metadata["total_pages"])
+        self.assertIsNotNone(metadata["total_posts"])
+        self.assertEqual(len(articles), 12)
+
+        # all other functionality is test via get_articles
```
tbille: ?

tests/test_wordpress_api.py
```diff
@@ -27,245 +20,181 @@ def raise_for_status(*args):
...
+
+        for article in articles_without_tags:
+            self.assertTrue(3278 not in article["tags"])
+
+        articles_all_have_one_tag, _ = api.get_articles(tags=[3278])
+
+        for article in articles_all_have_one_tag:
+            self.assertTrue(3278 in article["tags"])
+
+        # TODO: this is not yet implemented
```
tbille: Then remove the code

tests/test_wordpress_api.py
```diff
@@ -27,245 +20,181 @@ def raise_for_status(*args):
...
-        self.assertEqual(article, "hello_test")
-
-    @patch("canonicalwebteam.http.CachedSession.get")
-    def test_including_and_excluding_article(self, get):
-
-        get.return_value = MockResponse()
-
-        article = api.get_article(
-            slug="test", tags=[1234, 5678], tags_exclude=[9876]
+    @vcr.use_cassette("fixtures/vcr_cassettes/articles_with_metadata.yaml")
```
tbille: IMO you are testing way too much stuff. Don't test the API but your code.

What I want to know here is:

- Have you made an API call(s) when calling the function?
- What is happening if API replies 40x/50x?
- What happens if JSON broken?
- If there is a minimum of logic what are the edge cases?

tests/test_common_view_logic.py
b-m-f: The fixtures are all auto generated from the API call via https://vcrpy.readthedocs.io/en/latest/ .

But yes, the test relies on the library correctly creating the fixture.

I think breaking the app is a good idea. Ill try to add some of those.

tests/test_common_view_logic.py
b-m-f: This way the tests wont reflect the actual setup though. Id rather keep if for now. There shouldnt be any harm, should there?

If the cache misbehaves, the tests should fail :)

tests/test_common_view_logic.py
b-m-f: Im going to update them. But I do not agree with fully removing them, sorry.

tests/test_wordpress_api.py
b-m-f: Id rather keep the api test in as well, to make sure that regressions get caught.

For the error testing I have created https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/70

tests/test_wordpress_api.py
b-m-f: I have updated the comment to reflect the problem and created https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/67

tests/test_wordpress_api.py
b-m-f: Updated the comment

tests/test_common_view_logic.py
tbille: well then you are not reflecting the actual behaviour if you change the object. If you want to keep it then keep the Session as it is.

tests/test_common_view_logic.py
tbille: seems useless since code is quite self explanatory here. I am not confident with simple comments about their own maintenance.

tests/test_wordpress_api.py
tbille: same comment. Don't touch Session if you want to keep the same behaviour than prod.

tests/test_wordpress_api.py
tbille: This will be discussed together.

tests/test_wordpress_api.py
```diff
@@ -27,245 +20,182 @@ def raise_for_status(*args):
...
+        for article in articles_without_tags:
+            self.assertTrue(3278 not in article["tags"])
+
+        articles_all_have_one_tag, _ = api.get_articles(tags=[3278])
+
+        for article in articles_all_have_one_tag:
+            self.assertTrue(3278 in article["tags"])
+
+        # TODO:
+        # https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/67
```
tbille: this seems more something that you will be able to add directly when working on it.

tests/test_common_view_logic.py
```diff
@@ -1,389 +1,182 @@
...
+        )
+        article["categories"] = [1453]
+
+        complete_article = get_complete_article(article)
+        self.assertEqual(complete_article["author"]["name"], "Canonical")
+        self.assertEqual(
+            complete_article["image"]["id"], article["featured_media"]
+        )
+        self.assertEqual(complete_article["date"], "6 May 2019")
+        # Test for length of 339, and if > than 340, [...] is added
```
tbille: this is why I don't like comments. Sometimes they lie! :P
Where do you test this?
 
`self.assertLessEqual(len(complete_article["excerpt"]["raw"]), 344)` this just tells me that the raw is 344 characters.

What you want:

`self.assertEqual(complete_article["excerpt"]["raw"][-5:], "[...]")`

tests/test_wordpress_api.py
```diff
@@ -27,245 +20,182 @@ def raise_for_status(*args):
...
-    def test_including_and_excluding_article(self, get):
-
-        get.return_value = MockResponse()
-
-        article = api.get_article(
-            slug="test", tags=[1234, 5678], tags_exclude=[9876]
+    @vcr.use_cassette("fixtures/vcr_cassettes/articles_with_metadata.yaml")
+    def test_get_articles_with_metadata(self):
+        # This function is similar to get_articles, plus the metadata
+        # therefore all other functionality is tested via test_get_articles
```
tbille: no it's not similar.

You are getting the result of the function which is a correct test. This comment can be removed.

tests/test_common_view_logic.py
b-m-f: Hm, but what I want to test is that the amount of characters here is always less than 344, which is the theoretical maximum.

tbille: I agree with what you say but the comment tells me something else.

tests/test_wordpress_api.py
b-m-f: Could you explain this? I could still provide a myriad of parameters here, which @nottrobin already noticed, and needs a refactoring to stop using the `**kwargs` pass through to `build_get_articles_url`.

The comment is supposed to let the developer know that all the parameter testing is going to happen in the next test, and thus we need not replicate all of it here.

The only addition here is the metadata. I think this also shows that this part of the code could potentially needs some refactoring. F.e. add yet another parameter that enables metadata?

tests/test_common_view_logic.py
b-m-f: Unfortunately I cant test the "[...]" here since the article excerpt might be less than 340, in which case this is not added.

Rather than putting in an if here I think this calls for  a refactor of some sort, that encapsulates the trimming logic and can be tested in isolation.

tests/test_common_view_logic.py
b-m-f: Alright, Ill remove the timeout tolerance increase here to reflect the actual problems of the API.

b-m-f: Nvm, using `CachedSession` does not work without the timeout increase.

tests/test_common_view_logic.py
b-m-f: https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/70 

Created this to tackle the failing tests

tests/test_wordpress_api.py
```diff
@@ -26,247 +13,143 @@ def raise_for_status(*args):
...
-        article = api.get_article(slug="test", tags_exclude=[1234, 5678])
+class TestWordPressApi(unittest.TestCase):
+    @patch("canonicalwebteam.blog.wordpress_api.api_session.get")
+    def test_get_articles_with_metadata(self, get):
+        get.return_value = SuccessMockResponse()
+
+        articles, metadata = api.get_articles_with_metadata()
+        self.assertIsNotNone(metadata["total_pages"])
+        self.assertIsNotNone(metadata["total_posts"])
+        # test here that get was called with the correct url
```
tbille: I think the name of the function talks by itself here.

tests/test_common_view_logic.py
```diff
@@ -1,389 +1,174 @@
...
+        complete_article = get_complete_article(article)
+        self.assertEqual(complete_article["author"]["name"], "Canonical")
+        self.assertEqual(
+            complete_article["image"]["id"], article["featured_media"]
+        )
+        self.assertEqual(complete_article["date"], "6 May 2019")
+
+        # The maximum allowed character count for the excerpt is 339
+        # characters. If this limit is reached [...] is added
+        # which is why we test for a max length of 339 + 5 = 344 here
```
tbille: here we reach the limitations of the library. So this comment makes sense.

Maybe since it's an "edge" case we could unittest it. But not part of this PR anyway

README.md
```diff
@@ -85,3 +85,7 @@ The blog extension leverages [poetry](https://poetry.eustace.io/) for dependency
 
 ## Testing
 All tests can be run with `poetry run pytest`.
+
+### Regenerating Fixtures
+All API calls are caught with [VCR](https://vcrpy.readthedocs.io/en/latest/) and saved as fixtures in the `fixtures` directory. If the API should update, all fixtures can easily be updated by just removing the `fixtures` directory and rerunning the tests.
```
tbille: `s/If the API should update/If the API updates/`

README.md
```diff
@@ -85,3 +85,7 @@ The blog extension leverages [poetry](https://poetry.eustace.io/) for dependency
 
 ## Testing
 All tests can be run with `poetry run pytest`.
+
+### Regenerating Fixtures
+All API calls are caught with [VCR](https://vcrpy.readthedocs.io/en/latest/) and saved as fixtures in the `fixtures` directory. If the API should update, all fixtures can easily be updated by just removing the `fixtures` directory and rerunning the tests. 
+To do this run `rm -rf fixtures && poetry run pytest`.
```
tbille: Add line before to create a new paragraph, to give more "air" to the readme

PR #57, opened 2019-05-22
Updates the url building to be more maintainable

b-m-f:
# Done
Fixes https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/15
**Blocked** until https://github.com/canonical-web-and-design/canonicalwebteam.blog/pull/55 is merged and rebased into this branch

[urllib.parse](https://docs.python.org/3/library/urllib.parse.html ) was considered, but has the problem of returning a named tuple, which can not easily be updated.
In order to avoid misusage of the module this approach was decided against.
Instead some refactoring work is done to make URL building more maintainable, by keeping it almost exclusively in one function.

### DIFF
--- a/canonicalwebteam/blog/wordpress_api.py
+++ b/canonicalwebteam/blog/wordpress_api.py
@@ -19,10 +19,11 @@ def process_response(response):
     return response.json()
 
 
-def build_get_article_url(
+def build_url(
+    endpoint,
     tags=[],
-    per_page=12,
-    page=1,
+    per_page="",
+    page="",
     tags_exclude=[],
     exclude=[],
     categories=[],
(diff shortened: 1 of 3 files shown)


canonicalwebteam/blog/wordpress_api.py
```diff
@@ -31,43 +32,60 @@ def build_get_article_url(
...
+        "slug": slug,
+        "search": search,
+        "include": ",".join(str(id) for id in include),
+    }
+
+    # this will build the querystring with an appended &
+    # it does not matter to the API and avoids more logic
+    querystring = "?"
+    for param, value in params.items():
+        if value and value is not []:
```
tbille: from what I understand you have no lists anymore. So just `if value:` is enough

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -31,43 +32,60 @@ def build_get_article_url(
...
+        "sticky": sticky,
+        "before": before,
+        "after": after,
+        "slug": slug,
+        "search": search,
+        "include": ",".join(str(id) for id in include),
+    }
+
+    # this will build the querystring with an appended &
+    # it does not matter to the API and avoids more logic
```
tbille: please remove comment code should be self documented

PR #65, opened 2019-05-29
Only tries to parse the month variables if the data exists

b-m-f:
# Done
Updates an if statement to be more thorough

### DIFF
--- a/canonicalwebteam/blog/logic.py
+++ b/canonicalwebteam/blog/logic.py
@@ -97,13 +97,21 @@ def transform_article(
             article["content"]["rendered"]
         )
 
-    if "_start_month" in article:
+    if (
+        article.get("_start_month")
+        and article.get("_start_year")
+        and article.get("_start_day")
+    ):
         start_month_name = get_month_name(int(article["_start_month"]))
         article["start_date"] = "{} {} {}".format(
             article["_start_day"], start_month_name, article["_start_year"]
         )
 
-    if "_end_month" in article:
+    if (
+        article.get("_end_month")
+        and article.get("_end_year")
+        and article.get("_end_day")
+    ):
         end_month_name = get_month_name(int(article["_end_month"]))
         article["end_date"] = "{} {} {}".format(
             article["_end_day"], end_month_name, article["_end_year"]
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,4 @@
+1.5.15: Bugfix on date parsing in logic.py
 1.5.14: Updates upcoming articles to include additional data
 1.5.13: Adds filtering by multiple categories on archive view
 1.5.12: Add next 3 upcoming webinars and events to index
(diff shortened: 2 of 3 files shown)


canonicalwebteam/blog/logic.py
```diff
@@ -97,13 +97,15 @@ def transform_article(
             article["content"]["rendered"]
         )
 
-    if "_start_month" in article:
+    # if _start_month is set here we can assume that day and year
+    # will also have a value, same for the end data
+    if "_start_month" in article and article["_start_month"] != "":
```
jpmartinspt: Can this not just be article.get("start_month")?

b-m-f: Thanks! updated it to use `""` as a default value on get, to catch both the None and empty string scenario

canonicalwebteam/blog/logic.py
```diff
@@ -97,13 +97,15 @@ def transform_article(
             article["content"]["rendered"]
         )
 
-    if "_start_month" in article:
+    # if _start_month is set here we can assume that day and year
+    # will also have a value, same for the end data
+    if article.get("_start_month", "") != "":
```
tbille: no need for default value here.

canonicalwebteam/blog/logic.py
```diff
@@ -97,13 +97,15 @@ def transform_article(
             article["content"]["rendered"]
         )
 
-    if "_start_month" in article:
+    # if _start_month is set here we can assume that day and year
+    # will also have a value, same for the end data
```
tbille: no need to comment a if statement.

And assuming is not a good pattern

canonicalwebteam/blog/logic.py
tbille: ` if article.get("_start_month"):`

canonicalwebteam/blog/logic.py
```diff
@@ -97,13 +97,15 @@ def transform_article(
...
+    if article.get("_start_month", "") != "":
         start_month_name = get_month_name(int(article["_start_month"]))
         article["start_date"] = "{} {} {}".format(
             article["_start_day"], start_month_name, article["_start_year"]
         )
 
-    if "_end_month" in article:
+    if article.get("_end_month", "") != "":
```
tbille: same here

canonicalwebteam/blog/logic.py
b-m-f: Do you prefer a test on all attributes here?

This is not documented in the [wordpress Rest docs](https://developer.wordpress.org/rest-api/reference/posts/) as far as I can tell. Seems to be done with a plugin.

tbille: then what do you trust more?

b-m-f: Alright, I think this speaks for a full check then? Updated it.

PR #77, opened 2019-06-07
Refactor flask views

tbille:
# Summary

- Refactor flask views file to have the same shape as django
- Add missing endpoints (topic, author, upcoming, group, archives)


b-m-f:
LGTM, just need some update into the README for the new param on the blueprint :)

PR #86, opened 2019-06-17
Updates build_get_articles_url to use urllib.parse.urlencode

b-m-f:
# Done 
Fixes https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/15
Uses `urllib.parse.urlencode` in `build_get_articles_url`

# QA
Working tests reflect that the application still behaves normally


canonicalwebteam/blog/wordpress_api.py
```diff
@@ -47,24 +49,27 @@ def build_get_articles_url(
...
-    if before != "":
-        url = url + f"&before={before}"
-    if after != "":
-        url = url + f"&after={after}"
-
-    return url
+    params_object = [
+        param
+        for param in [
+            ("tags", ",".join(str(id) for id in tags)),
```
nottrobin: Remind me, what's the expected function of `tags`? Is it supposed to select:

1. Topics with any of the listed tags (union)? or:
2. Topics with all of the listed tags (intersection)?

b-m-f: 2. Intersection :)

nottrobin: Could you clarify this in the function description?

But speaking of that, I don't see why this function needs to be specific, as we discussed before. I thought we were going to make it so this could be a generic URL builder. You treat all arrays the same way - you turn them into comma-separated lists. So it seems to me you really just want a function that turns a dictionary of items into a wordpress-style API URL.

And then, as we discussed before, you can move the explicit arguments into the definition of `get_articles`.

b-m-f: Ah, alright. Couldnt remember from the ticket. Updated it accordingly.

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -19,71 +21,64 @@ def process_response(response):
...
-        f"&exclude={','.join(str(id) for id in exclude)}"
-        f"&author={author}&_embed"
-    )
-    if sticky != "":
-        url = url + f"&sticky={sticky}"
-    if before != "":
-        url = url + f"&before={before}"
-    if after != "":
-        url = url + f"&after={after}"
+    params_object = [param for param in params.items() if param[1]]
```
nottrobin: I don't think this does what you want:

``` python3
In [2]: params = {'tags': [1, 2, 3, 4]}

In [3]: params_object = [param for param in params.items() if param[1]]

In [4]: params_object
Out[4]: [('tags', [1, 2, 3, 4])]

In [5]: from urllib.parse import urlencode

In [6]: from urllib.parse import urlencode, unquote

In [7]: unquote(urlencode(params_object))
Out[7]: 'tags=[1,+2,+3,+4]'
```

I think what you want is more along the lines of:

``` python3
for key, value in params:
    if type(value) is list:
        params[key] = ",".join([str(item) for item in value])

query = urlencode(params_object)
```

b-m-f: Oh yeah! Saves a lot of repition. 
Ive adjusted it slightly to only run on parameters that actually have values.

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -19,71 +21,74 @@ def process_response(response):
...
-    )
-    if sticky != "":
-        url = url + f"&sticky={sticky}"
-    if before != "":
-        url = url + f"&before={before}"
-    if after != "":
-        url = url + f"&after={after}"
 
-    return url
+    clean_params = [(key, value) for key, value in params.items() if value]
```
nottrobin: You do this to filter out items whose value is `None` or `False`? Is this actually a case you're dealing with, and are you sure that's what you want to do? This seems to add a lot of complexity.

I would have thought you'd expect people using this function to know what they're doing and be passing through each parameter deliberately, they wouldn't expect some parameters to be filtered out. So either, do something special with `None` or `False`, or just do nothing and let them be converted into strings naturally. What you're doing here feels like over-engineering for a case you probably don't have.

Even if you did want to do this, it's a shame to convert a relatively easy to use `dict` into a confusing list of tuples. You should keep it as a dictionary which would be much easier to work with in your following for loop. You could do this through a dictionary comprehension:

``` python3
clean_params = {key: value for key, value in params.items() if value}
```

But as I say, in this case I'd probably suggest not doing it at all.

b-m-f: Hm yeah, I was on this for a bit. The problem with explicitely matching `None` is that I wont be able to exclude all the empty strings.

Ill use lamda funtions here. They dont  flow so well in Python but should solve this nicely.

b-m-f: Updated it. 

I feel like funtional programming is just not well supported, but the flow seems much cleaner here.

Would love to chain the function without nesting in parantheses. Is there a way to do it in Python?

canonicalwebteam/blog/common_view_logic.py
```diff
@@ -42,13 +42,14 @@ def get_index(self, page=1, category="", enable_upcoming=True):
...
 
-        articles, total_pages = api.get_articles(
+        articles, metadata = api.get_articles(
             tags=self.tag_ids,
             tags_exclude=self.excluded_tags,
             exclude=[article["id"] for article in featured_articles],
             page=page,
             categories=[category_id],
         )
+        total_pages = metadata["total_pages"]
```
tbille: this is done twice (also in the get_index_context if I am nto wrong)

b-m-f: on line 57 the `total_pages` gets passed into  `get_index_context`. Shouldnt happen twice :)

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -19,71 +21,78 @@ def process_response(response):
...
+
+    def list_to_comma_separated_string(key, value):
+        if type(value) is list:
+            return (key, (",".join(str(item) for item in value)))
+        else:
+            return (key, value)
+
+    params = {key: value for key, value in params.items() if value}
+    params = dict(
+        map(lambda kv: list_to_comma_separated_string(*kv), params.items())
```
nottrobin: I still feel you're over-engineering this. Why isn't a simple `for key, value in params.items()` sufficient? Lambdas, inline functions, lists of tuples and dictionary comprehensions are just fundamentally confusing constructs for people to read when a much more simple construct would work just as well.

E.g., in this case how is what you've written any different from what I previously suggested?

``` python3
for key, value in params.items()
    if type(value) is list:
        params[key] = ",".join(str(item) for item in value)
    elif not value:
        del params[key]
```

I also would rather that you were more explicit about what sorts of items are going to get removed from the dictionary (rather than simply `not value`), as I personally find that quite a confusing behaviour.

For exampe, you mentioned empty strings. I would expect that if I did something like `build_query({"a": "1", "b": False, "c": ""})` then what I'd get would be `"a=1&b=false&c`. I understand that for your specific use-case in `get_articles` you want to be able to pass through `None` for some of the values and have them not end up in the built query string, but I would keep it to an explicit behaviour for `None`, and make sure you document it in the function definition.

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -19,71 +21,78 @@ def process_response(response):
...
+    )
 
 
-def get_articles_with_metadata(**kwargs):
+def get_articles(
+    tags=[],
+    tags_exclude=[],
+    exclude=[],
+    categories=[],
+    sticky="",
```
nottrobin: It's common practice to set optional arguments to `None`, unless you actually intend to use a different value. So unless you intend to have `sticky=""` in your resulting API query, I think `None` would be a lot more intuitive.

b-m-f: Aight. Changing this makes the `for loop` approach easy.

canonicalwebteam/blog/wordpress_api.py
b-m-f: Aight. Updated it.

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -19,71 +21,74 @@ def process_response(response):
...
 
-    return url
+    clean_params = {}
+    for key, value in params.items():
+        if type(value) is list:
+            clean_params[key] = ",".join(str(item) for item in value)
+        elif value is not None:
+            clean_params[key] = value
+
+    params = urlencode(clean_params)
```
nottrobin: I think maybe you meant to call this variable `query` or something

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -19,71 +21,74 @@ def process_response(response):
...
-    if sticky != "":
-        url = url + f"&sticky={sticky}"
-    if before != "":
-        url = url + f"&before={before}"
-    if after != "":
-        url = url + f"&after={after}"
 
-    return url
+    clean_params = {}
+    for key, value in params.items():
```
nottrobin: Thanks :heart:

canonicalwebteam/blog/wordpress_api.py
b-m-f: yeah, Much better name!

PR #87, opened 2019-06-18
Add pagination for author endpoint

tbille:
# Summary

Add pagination for author endpoint.

### DIFF
--- a/canonicalwebteam/blog/django/views.py
+++ b/canonicalwebteam/blog/django/views.py
@@ -62,8 +62,9 @@ def upcoming(request):
 
 
 def author(request, username):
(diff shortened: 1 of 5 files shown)


canonicalwebteam/blog/django/views.py
```diff
@@ -62,8 +62,9 @@ def upcoming(request):
 
 
 def author(request, username):
+    page_param = request.GET.get("page", default="1")
```
b-m-f: Will this crash when inserting `abc` for page?

tbille: Opened an issue for it: https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/88

PR #89, opened 2019-06-18
Add number of posts on author endpoint

tbille:


tbille:
updated @anthonydillon 	thanks for the quick review

PR #95, opened 2019-06-27
Fix archives 500

sowasred2012:
## Done
- Set `after` and `before` default values to `None` in the `get_archives` method.
  - The Wordpress API returns an error if the `after` and `before` values are empty strings, which was the case unless a `year` param was present in the URL.

## QA
- Use https://github.com/canonical-web-and-design/kubeflow-news.com and replace `canonicalwebteam.blog==2.0.x` with `git+https://github.com/sowasred2012/canonicalwebteam.blog@fix-archives-500#egg=canonicalwebteam.blog` in `requirements.txt`
- Run `./run clean && ./run`
- Visit [/archives](http://0.0.0.0:8035/archives)
- Ensure that you see the archives page, rather than an error

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,4 @@
+2.0.10: Fix get_archives error caused by default after/before params being empty strings
 2.0.4: Internal API refactor
 2.0.3: Add pagination for author endpoint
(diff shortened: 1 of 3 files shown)


nottrobin:
logic looks right to me

PR #110, opened 2019-08-07
Use uncached sessions for feed API calls

### LINKED ISSUE #111: Don't cache feeds responses
Create a separate requests session for feeds requests, that does not cache the responses, in order to avoid the cache directory growing in size.

jpmartinspt:
Add new `UncachedSession` to use for feeds related requests.

QA
- Run the tests for the project.

Fixes #111


nottrobin:
Nice and simple. But I feel we should test this a little bit before we actually make it the next official release.

Can we do some kind of benchmarking of ubuntu.com/blog with this versions vs the old version?

jpmartinspt:
I'm trying it locally on ubuntu.com. I can try to run some benchmarks.

PR #113, opened 2019-08-15
Update 'Getting started' docs in README

barrymcgee:
## Done

Update 'Getting started' docs in README based on my recent usage of this module on jaas.ai

## QA

- Please review my changes carefully, I'm not an expert in Python, much less extension development. 
- Check my attempt to explain the params this extension will accept and how I've attempted to document them. Are the descriptors of params correct, including types? Is there a better way to describe them?

Fixes https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/104

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Ubuntu blog extension
+# Canonical blog extension
 
 This extension allows you to add a simple frontend section to your flask app. All the articles
 are pulled from the (Ubuntu blog)[https://ubuntu.com/blog] WordPress API using JSON.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Canonical blog extension
 
 This extension allows you to add a simple frontend section to your flask app. All the articles
-are pulled from the (Ubuntu blog)[https://ubuntu.com/blog] WordPress API using JSON.
+are pulled from `[Canonical's Wordpress back-end](https://admin.insights.ubuntu.com/wp-admin/)` through the JSON API.
 
 This extension provides a blueprint with 3 routes:
 
--- a/README.md
+++ b/README.md
@@ -1,38 +1,69 @@
-# Blog flask extension
+# Ubuntu blog extension
 
-This extension allows you to add a simple blog frontend to your flask app. All the articles
-are pulled from the WordPress API that has the plugin WP-JSON.
+This extension allows you to add a simple frontend section to your flask app. All the articles
+are pulled from the (Ubuntu blog)[https://ubuntu.com/blog] WordPress API using JSON.
 
 This extension provides a blueprint with 3 routes:
+
 - "/": that returns the list of articles
 - "/<slug>": the article page
 - "/feed": provides a RSS feed for the page.
+	
+## How to install
+
+To install this extension as a requirement in your project, you can use PIP;
+
+```bash
+pip install canonicalwebteam.blog
+```
+
+See also the documentation for (pip install)[https://pip.pypa.io/en/stable/reference/pip_install/].
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,38 +1,69 @@
...
+
 ### Django
 
 - Add the blog module as a dependency to your Django project
-- Load it at the desired path (f.e. "/blog") in the `urls.py` file
+- Load it at the desired path (e.g. "/blog") in the `urls.py` file
+
 ```python
 from django.urls import path, include
-urlpatterns = [path(r"blog/", include("canonicalwebteam.blog.django.urls"))]
```
barrymcgee: This was a typo, right?

nottrobin: The `r`? `r` means ["raw string"](https://www.journaldev.com/23598/python-raw-string), which is a pattern often used for URL routes, which often contain regular expressions - so it avoids needing to escape regex control characters. I don't have a strong opinion on whether to use `r` here or not.

README.md
```diff
@@ -1,38 +1,69 @@
-# Blog flask extension
+# Ubuntu blog extension
```
nottrobin: I don't quite agree with this characterisation. This is not just for ubuntu.com/blog, but also for snapcraft.io/blog, jaas.ai/blog (as you know) etc.. And it's quite possible that we might in the future make it so snapcraft blog posts don't show up at ubuntu.com/blog at all.

I consider ubuntu.com/blog to be just one implementation of this module.

It is, however, currently inherently true that this blog module gets its data from the Wordpress at admin.insights.ubuntu.com. So I wouldn't mind mentioning that fact in the description somewhere. However, it does seem neatest to me to continue to describe the module itself as simply "Blog" or maybe "Canonical blog".

README.md
barrymcgee: I'll revert it.

README.md
barrymcgee: This isn't attempting to describe where this module is used, but the hard-coded feed it will output. 

When I first came to this module, I had no idea the feed was hard-coded and spent quite a while trying to figure out how to pass it a JSON endpoint. The feed comes from `https://admin.insights.ubuntu.com`, the front-end of which used to be `https://insights.ubuntu.com` which now redirects to `https://ubuntu.com/blog` - therefore it's not a generic extension to output a blog feed, it can only output posts from the ubuntu blog wordpress installation, therefore that should be made clear immediately in the title.

cristinadresch:
From stand-up

README.md
nottrobin: We can call it "Canonical blog" if you like.

Regardless of its history, what it actually is is a plugin that pulls content from a CMS API back-end that is the only place Canonical as a company publish blog posts to, whether those blog posts be about Ubuntu, Snapcraft, Jaas or anything else that Canonical get up to. And it's not a place that the Ubuntu community can publish to. So it isn't "Ubuntu blog".

barrymcgee: Done

README.md
```diff
@@ -1,38 +1,69 @@
-# Blog flask extension
+# Canonical blog extension
 
-This extension allows you to add a simple blog frontend to your flask app. All the articles
-are pulled from the WordPress API that has the plugin WP-JSON.
+This extension allows you to add a simple frontend section to your flask app. All the articles
+are pulled from the (Ubuntu blog)[https://ubuntu.com/blog] WordPress API using JSON.
```
nottrobin: This should be something like:

> are pulled from `[Canonical's wordpress back-end](https://admin.insights.ubuntu.com/wp-admin/)` through the JSON API.

PR #114, opened 2019-08-15
Add feeds endpoints for authors, groups and topics

jpmartinspt:
## DONE
Add missing feeds endpoints and build the feed ourselves instead of redirecting to WP.

## TODO
- [x] Have a better look at URL generation
- [x] Update feed titles for author, groups and topics.

## QA
- Pull this branch.
- Run `poetry build`
- In your fork of `www.ubuntu.com` pull this branch https://github.com/canonical-web-and-design/ubuntu.com/pull/5693/files.
- In `www.ubuntu.com` requirements.txt swap the `canonicalwebteam.blog` requirement for the local version you cloned on the first step(use the tar.gz in the `dist` folder of your local copy of `canonicalwebteam.blog`).
- Run without docker and try to append `/feed` to different blog url's (groups, topics and authors) and make sure it returns similar content to what you see on the actual page.
- Do the same on a flask project like `jp.ubuntu.com` (no need to clone any specific branch, just swap the requirement)


canonicalwebteam/blog/django/views.py
```diff
@@ -55,6 +76,21 @@ def group(request, slug, template_path):
...
+    try:
+        feed = blog_views.get_group_feed(
+            group_slug=slug,
+            uri=request.build_absolute_uri(),
+            title=title,
+            subtitle=subtitle,
+            article_url_generator=article_url_generator,
+        )
+    except Exception as e:
+        return HttpResponse("Error: " + str(e), status=502)
```
nottrobin: I know you're just following precedent here, but I really dislike this pattern. I assume this is just so that we pass a `502` rather than a `500` status? But I really don't think we necessarily know if that's accurate. The error could be coming from anywhere, in our code, or whatever. I think we should only return e.g. `502` in really specific circumstances where we definitely know that it's wordpress's fault.

Would you mind not adding more of these `try..except` blocks in new functions? Just let errors bubble up to be handled by the application as normal please.

jpmartinspt: I agree and would like to remove it in all views, since this is actually masking possible errors(we don't get them in sentry). I can do this now for this views, or do them all in a separate PR?

nottrobin: Or do them all in this PR? I don't mind, up to you, but please just don't add more of them =)

canonicalwebteam/blog/wordpress_api.py
```diff
@@ -8,14 +8,15 @@
...
 
 api_session = CachedSession(
     fallback_cache_duration=300, file_cache_directory=".webcache_blog"
 )
+uncached_api_session = UncachedSession()
+
 
-feed_session = UncachedSession()
+def _get_api_session(cache=True):
+    return api_session if cache else uncached_api_session
```
nottrobin: Isn't this a bit trivial for its own function?

canonicalwebteam/blog/feeds.py
```diff
@@ -0,0 +1,46 @@
...
+    title = article["title"]["rendered"]
+    author = article["_embedded"]["author"][0]
+    description = article["excerpt"]["rendered"]
+    content = article["content"]["rendered"]
+    published = f'{article["date_gmt"]} GMT'
+    updated = f'{article["modified_gmt"]} GMT'
+
+    # TODO: Improve URL generation
+    host = uri.split("/")[2]
+    path = article_url_generator(slug=article["slug"])
```
nottrobin: Couldn't we just do `path = article['slug']`? I'm really not sure passing a `article_url_generator` function through 2 levels of encapsulation just to resolve to `/{slug}` is worth it.

In general, I'm not a fan of `url_for` patterns, as in almost every case anyone reading that line just wants to know what the eventual URL looks like, and working that out from a `url_for` function always involves multiple steps. I'm not sure I've ever seen anyone benefit from the theoretical upside of being able to change the URL in just one place (urls.py).

nottrobin:
Apart from my few pernickerty comments about specific code structure, this looks exactly the right sort of shape. Thanks @jpmartinspt. Presumably it's not too far from being complete?

canonicalwebteam/blog/feeds.py
nottrobin: Or is this needed because you don't know the subpath? (`/blog` or not)

PR #117, opened 2019-08-22
Order feeds from most recent to oldest

jpmartinspt:
# Done
Show most recent entries first in feeds endpoints.

# QA
- In `www.ubuntu.com` replace the `canonicalwebteam.blog` dependency with this branch.
- Go to feeds endpoints and make sure the most recent entries come first.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "canonicalwebteam.blog"
-version = "2.4.0"
+version = "2.4.1"
 description = "Flask extension and Django App to add a nice blog to your website"
 authors = ["Canonical webteam <webteam@canonical.com>"]
 
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
     name='canonicalwebteam.blog',  # Required
     # https://www.python.org/dev/peps/pep-0440/
     # https://packaging.python.org/en/latest/single_source_version.html
-    version='2.4.0',  # Required
+    version='2.4.1',  # Required
     # https://packaging.python.org/specifications/core-metadata/#summary
     description="Flask extension and Django App to add a nice blog to your website",  # Required
     # https://packaging.python.org/specifications/core-metadata/#description-optional
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,4 @@
+2.4.1: List feeds from most recent to oldest
 2.4.0: Add feeds endpoints for groups, topics and authors
 2.3.0: Use uncached sessions for API calls
 2.2.1: Fix up default Title setting for blog feed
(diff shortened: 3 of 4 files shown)


requirements.txt
```diff
@@ -1,7 +1,9 @@
 # IMPORTANT: this file is autogenerated. Do not edit it manually.
+Flask (>=1.0,<2.0)
```
nottrobin: Double flask?

setup.py
```diff
@@ -46,6 +46,8 @@
     packages=find_packages(),  # Required
     # https://packaging.python.org/en/latest/requirements.html
     install_requires=[
+        'django[django] (>=2.2,<3.0)',
```
nottrobin: This is a funny dependency

requirements.txt
```diff
@@ -1,7 +1,9 @@
...
 Flask (>=1.0.1,<2.0.0)
 canonicalwebteam.http (>=1.0.1,<2.0.0)
 django (>=2.2,<3.0)
+django[django] (>=2.2,<3.0)
```
nottrobin: Double Django?

setup.py
```diff
@@ -46,6 +46,8 @@
...
+        'Flask (>=1.0,<2.0)',
```
nottrobin: These should be under an `extras_require` section

pyproject.toml
```diff
@@ -13,16 +13,14 @@ license = "LGPL-3.0"
 
 [tool.poetry.dependencies]
 python = "^3.6"
-django = {version = "^2.2", extras = ["django"]}
+django = {version = "^2.2", extra = ["django"]}
```
nottrobin: Actually, I think `extras` is the right word. And it should be `extras` for Flask too.

nottrobin: At least that's what the example here suggests: https://github.com/sdispater/poetry#introduction

jpmartinspt: `poetry install` fails with `extras`.

nottrobin: yeah that's cos it's right =) what we should actually be doing is `optional = True` or something I think

nottrobin: https://poetry.eustace.io/docs/pyproject/#extras

nottrobin: But the simplest thing to do would just be to purely update the version in these files for the time being. None of this stuff is related to this PR.

jpmartinspt: Yup, but that then fails for `poetry-setup` :man_shrugging:

nottrobin: But FWIW, I think the right syntax should be:

``` toml
django = {version = "^2.2", optional = true}
flask = {version = "^1.0", optional = true}

[tool.poetry.extras]
django = ["django
flask = ["flask"]
```

nottrobin: Yeah but you don't *need* to fix or run `poetry-setup` for this PR do you? You can literally just update the number in `setup.py` and `pyproject.toml`

jpmartinspt: Yeah, done that now.

nottrobin:
The diff now looks good, but it also looks like you're trying to overwrite history in this PR:

image

Could you separate your work out into a new commit?

jpmartinspt:
Fixed now.

PR #132, opened 2020-01-13
Tests and publish package with GitHub Actions

jkfran:
### Done
- Remove CircleCI config
- Build check for new PRs
- Build package and publish to PyPI when the branch `master` is updated

### QA
Make sure the checks with the "Tests /" prefix pass

### Issue
https://github.com/canonical-web-and-design/web-squad/issues/2094


pyproject.toml
```diff
@@ -14,7 +14,7 @@ license = "LGPL-3.0"
 [tool.poetry.dependencies]
 python = "^3.6"
 django = {version = "^2.2", extras = ["django"]}
-Flask = {version = "^1.0", extra = ["flask"]}
+Flask = {version = "^1.0"}
```
nottrobin: ?

jkfran: @nottrobin The build was failing at this line, maybe because "extra" should be "extras". But anyway this is redundant, right?

nottrobin: I don't think so... or if it is then it's written wrong.

What this is trying to say is that if you're installing `blog` for Django then you do it with `canonicalwebteam.blog[django]`, if for Flask, you do it with `canonicalwebteam.blog[flask]`. If this is working as it should be, your change would mean you'll be installing Flask even if you're trying to use the module with Django.

*But* I don't know if anything still uses the blog with Django, or if anything does (cn.ubuntu.com maybe?), maybe we should be converting that project to Flask. Which would mean we could release a new major version of this module that removed Django support and makes Flask mandatory.

jkfran: I could be wrong but then I think we are not doing it properly. Basically, if we want to do that theses dependencies should be optional = true and then we need to add a section to define extras, take a look:
https://python-poetry.org/docs/pyproject/#extras

So what we this library have been doing means something totally different, it means install Flask or django with the extra "flask" or "django" from those libraries. I just did this small test to check it:
`poetry add splinter -E django`
If I am right it is the equivalent to `pip install 'splinter[django]'` because:
```
-E, --extras=EXTRAS             Extras to activate for the dependency. (multiple values allowed)
```
And now in the pyproject.toml I had:
```
splinter = {version = "^0.13.0", extras = ["django"]}
```

nottrobin: Does that mean that every Flask project we have added `canonicalwebteam.blog` to is also installing `Django`? 'cos that's what I would expect from what you're saying. That's a big deal.

jkfran: Looks like we probably are

nottrobin: I'll verify that when I get a sec

nottrobin: God damn you're right

nottrobin: I filed: https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/133

PR #138, opened 2020-06-09
v6: Mahoosive restructure

nottrobin:
- Put API functions in a class object (Wordpress)
- Rewrite tests to use VCR and remove tests of internal functions
- Improve arguments to build_feed function
- Remove Django files
- Update build_blueprint arguments a bit
- Rename common_view_logic.py to views.py
- Rename logic.py to helpers.py

QA
--

I've implemented this in https://github.com/canonical-web-and-design/ubuntu.com/pull/7822, so follow the QA steps in that PR to test this module.


jkfran:
We should update also the PR checks, like:
https://github.com/canonical-web-and-design/canonicalwebteam.blog/blob/master/.github/workflows/tests.yml#L20

I am not sure why they are not running in this PR anyway.

nottrobin:
This is WIP until I get around to creating a branch on ubuntu.com that makes use of this new version  - until then it can't be QAed properly.

PR #140, opened 2020-07-07
Change image URL

albertkol:
Find and replace any references to https://admin.insights.ubuntu.com/wp-content with https://ubuntu.com/wp-content.

Make the link replacement optional.

QA
--
1. Go to your local ubuntu.com repo. Enable the new blog changes for testing:
```bash
git remote add albert-fork git@github.com:Albert-Arthur/ubuntu.com.git  # get my fork
git fetch albert-fork  # fetch the branches
git checkout enable-transform-link  # checkout the branch

dotrun  # run website
```

Browse to `/blog`

The thumbnails and images should work as usual, and if you inspect element you will see that
the links point to `ubuntu.com/wp-content` instead of `admin.insights.ubuntu.com`

Similarly if you go to a blog page fx:
`/blog/dell-xps-13-developer-edition-with-ubuntu-20-04-lts-pre-installed-is-now-available`

The image will again point to `ubuntu.com/wp-content`.

## Fixes #2130 step 2


canonicalwebteam/blog/ub_wordpress.py
```diff
@@ -0,0 +1,52 @@
+from canonicalwebteam.blog import Wordpress
+from canonicalwebteam.blog.helpers import transform_article
+
+
+class UbWordpress(Wordpress):
```
jpmartinspt, line 5: Could the current `Wordpress` class just take a `transform_article` function? Sounds easier than creating a new class just to return transformed articles, in my opinion.

canonicalwebteam/blog/helpers.py
```diff
@@ -128,22 +126,33 @@ def transform_article(article, group=None):
...
 
+    if transform_links:
+        article = recursive_search_and_replace(article)
+
     return article
 
 
-def change_url(feed, host):
-    """Change insights urls to <host>/blog
+def recursive_search_and_replace(article):
```
jpmartinspt, line 135: Do we need recursion and type checking here? I would expect we always call `transform_article` on a single article of the same type, so we should only need to replace the links on `article["content"]`(not sure that is the right key)

canonicalwebteam/blog/ub_wordpress.py
jpmartinspt: On another note if we are to use a class, can da name be a bit more descriptive? I don't know what `UbWordpress` means(I might do, but that's the point 😄)

nottrobin: @jpmartinspt: @albertkol and I discussed this a bit. Your suggestion is one option, but it breaks the purity of the `Wordpress` class being a straightforward representation of the data from the Wordpress API. We discussed this alternate structure as a way to keep the opinionated code that does data transformations for our own purposes separate from that pure class.

I have my own notes along those lines:

I'm not a fan of abbreviated variable names, and I *think* they are considered unpythonic (though they're not explicitly mentioned in PEP8 I don't think).

In any case, my thinking with this class is that what separates it from `Wordpress` is that instead of being a representation of a Wordpress API, it's instead an opinionated interface for retrieving Blog data. So ideally the fact that it's for a Wordpress API backend could now be abstracted away from the user. So maybe we could call it `BlogAPI` or something?

canonicalwebteam/blog/ub_wordpress.py
```diff
@@ -0,0 +1,52 @@
...
+            before,
+            after,
+            author,
+            groups,
+            per_page,
+            page,
+        )
+
+        return (
+            [transform_article(a, self.transform_links) for a in articles],
```
nottrobin, line 45: `transform_article` is really the whole point of this class - it's basically the only thing that differentiates it from `Wordpress` right now. Could we move that helper function inside this class instead? `def _transform_article`?

canonicalwebteam/blog/ub_wordpress.py
jpmartinspt: @nottrobin that's a good point, but if the only way to customize transformations is by extending `Wordpress` we are still not abstracting people away from WP. Maybe `transform_article` should be passed down to `BlogViews` since it's the view's responsability to give the templates tha right context to render? We can also just default the behaviour to returning the article as is and possibly leave the implementation of the `transform_article` specific to each project that needs to transform articles?

```python3
class BlogViews:
    def __init__(
        self,
        api,
        tag_ids=[],
        excluded_tags=[],
        blog_title="Blog",
        blog_path="blog",
        feed_description=None,
        per_page=12,
        transform_article=lambda article: article
    ):
        self.api = api
        self.tag_ids = tag_ids
        self.excluded_tags = excluded_tags
        self.blog_title = blog_title
        self.blog_path = blog_path.strip("/")
        self.feed_description = feed_description or f"{blog_title} feed"
        self.per_page = per_page
        self.transform_article = transform_article
```

nottrobin: Yeah, I discussed this a bit with @albertkol as well I think.

There's always a choice to be made between encoding opinionated functionality into a module, making the module inflexible, or making the module generic, to be configured from the outside, which will mean that each user of the module will configure it their own way rather than using a standard implementation.

In general, I think it's a better idea for modules to always be opinionated about what the right single solution is, and then only build in the ability to override this functionality when that need becomes clear. This helps the module remain simple and straightforward to use. This movement / philosophy is broadly called [convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration) I believe.

So, I think we should at least be making sure the module uses our `transform_article` function *by default*, than the question becomes, should we make that function overrideable?

As I see the point of our modules, it is to be deliberately opinionated so we are rolling our standard implementations to all our sites. This concern is generally much more important than the potential needs of the community. We don't focus our time and attention on making community modules, but instead on building our websites, and we open it up in case it can be of interest or use to the public, not because we write it for them. So I would say we want to lean strongly on the side of having strong opinions and not make much configurable until there's any actual need.

So the question of whether we should allow the client to pass through an override for the default `transform_article` function, for me, comes down to whether that would be any use to @mmallejac. My interpretation of what he asked for was simply to be able to turn off the `cloudinary.com` prefix, although as far as I understand it, this was only ever in a template, which should be easily overrideable.

I think we should not provide the ability to pass through this function until we know someone legitimately needs it.

albertkol: As a result of the whole conversation above.. 

I will rename the `UbWordpress` to `BlogAPI`.

For now the article transformation will not allow to be overwritten. If we choose to reconsider it in the future, I'd gladly make the change.

canonicalwebteam/blog/helpers.py
albertkol: I removed the recursive replace. Instead there are two places I am replacing the url manually:
- `article['content']['rendered']` for the article page content
- `article['image']['source_url']` for the thumbnails

It was an option I considered initially. I thought recursivity will cover some places I am unaware of. But it should be fine with these two.

canonicalwebteam/blog/ub_wordpress.py
albertkol: I moved `transform_article` inside the class. I moved the rest of the functions from the `helpers` file that were used only by `transform_article` too. 

I deleted the functions from the `helpers` file that were unused. 

In the end I was left with two functions which I moved to the `BlogViews` class, which allows me to get rid of the `helpers` file completely.

README.md
```diff
@@ -60,11 +60,21 @@ blog = build_blueprint(
         exclude_tags=[26, 34],
         per_page=12,
         feed_description="The Ubuntu Blog Feed",
-        api=Wordpress(session=session),
+        api=BlogAPI(session=session),
     )
 )
 ```
 
+If you wish to get the raw format of the articles use the `Wordpress` class instead of `BlogAPI`:
```
nottrobin, line 68: I dunno if you need to mention this. I think there's only 1 recommended way to set up this module and that's with the `BlogAPI`. The `Wordpress` class is just for purity - in that we could extract that class for use somewhere else if we wanted.

I don't feel strongly, but this seems to complicate the README for little utility.

PR #142, opened 2020-07-15
Use image_template for image processing

### LINKED ISSUE #139: Cloudinary
Hello team

Using another Wordpress instance is easy with the help of the BLOG_API env. variable.

But I haven't found something similar for Ubuntu's cloudinary image caching mechanism.
It is hardcoded within `logic.py` : `replace_images_with_cloudinary`

Any suggestion to override this ? 

Many thanks,
Michel

albertkol:
Enable cloudinary again for our blog articles.

This can be turned on or off with the flag `use_image_template=False` during the blog initialisation. 

QA
--
Browse to a blog article page that contains an image fx:
https://ubuntu-com-7940.demos.haus/blog/dell-xps-13-developer-edition-with-ubuntu-20-04-lts-pre-installed-is-now-available

or

https://ubuntu-com-7940.demos.haus/blog/design-and-web-team-summary-20th-july-2020

Check out the image url, it will be converted to a cloudinary url.

## Fixes #139


tests/test_blueprint.py
```diff
@@ -56,6 +56,7 @@ def test_article_with_image(self):
             b"admin.insights.ubuntu.com/wp-content/uploads", response.data
         )
         self.assertIn(b"ubuntu.com/wp-content/uploads", response.data)
+        self.assertIn(b"cloudinary", response.data)
```
jpmartinspt, line 59: Could we test it actually does the transfromation we expect?

tests/test_blog_api.py
```diff
@@ -48,3 +49,5 @@ def test_get_articles_without_transforming_links(self):
             "admin.insights.ubuntu.com/wp-content/uploads",
             article["content"]["rendered"],
         )
+
+        self.assertNotIn("cloudinary", article["content"]["rendered"])
```
jpmartinspt, line 53: Not sure about this assert here. There is no link to transform, but `cloudinary` could still be in the article content even if there are no trnasformed links.

tests/test_blog_api.py
```diff
@@ -35,6 +35,7 @@ def test_get_articles_with_transforming_links(self):
         self.assertIn(
             "ubuntu.com/wp-content/uploads", article["content"]["rendered"]
         )
+        self.assertIn("cloudinary", article["content"]["rendered"])
```
jpmartinspt, line 38: Could we test it actually does the transfromation we expect?

jpmartinspt:
Works well, code is good, just some comments on the tests.

tests/test_blog_api.py
nottrobin: Yes, given that the tests should be running against fixtures anyway, we should be able to pick out a specific image URL in that article that we expect to have been, or not to have been, transformed, and check the state of that specific URL.

nottrobin:
A question - does this only change images in the body of the article? What about the thumbnails on the index page?

I know that at the moment in e.g. `jp.ubuntu.com` the cloudinary prefix is added inside the template, but this isn't ideal. Really, if this module is going to do the transformation anywhere it should do it everywhere, and the prefix should be removed from templates.

That also therefore means that you need to make this at least another minor version change, as it effects the module's API.

canonicalwebteam/blog/blog_api.py
```diff
@@ -135,6 +135,11 @@ def _transform_article(self, article):
                 article["content"]["rendered"]
             )
 
+            # use cloudinary for blog article images
+            article["content"]["rendered"] = self._replace_img_to_cloudinary(
```
nottrobin, line 139: Aren't you gonna make this optional through a parameter somewhere? Otherwise this kinda does the exact opposite of what is asked for in https://github.com/canonical-web-and-design/canonicalwebteam.blog/issues/139

albertkol:
> A question - does this only change images in the body of the article? What about the thumbnails on the index page?
> 
> I know that at the moment in e.g. `jp.ubuntu.com` the cloudinary prefix is added to thumbnails inside the template, but this isn't ideal. Really, if this module is going to do the transformation anywhere it should do it everywhere, and the prefix should be removed from templates.
> 
> That also therefore means that you need to make this at least another minor version change, as it effects the module's API.

I guess I thought it should be the website's responsibility to handle the processing of the thumbnails and how it does it. Please don't ask me why... 

It makes sense to have the module cover it.

I gave it a thought. I might have an idea how to make the module cover thumbnails too.

canonicalwebteam/blog/blog_api.py
albertkol: It is optional under the same parameter `self.transform_links` which decides the admin.insights url rewrites.

I should have created a new one.

nottrobin: No it's ok for it to be both I think. But the `admin.insights.ubuntu.com` one doesn't actually matter that much, maybe that doesn't even need to be optional - because no-one else will be using that domain, and we will *always* want to replace it I think.

tests/test_blueprint.py
albertkol: I changed all tests mentioned to check for the `src` transformation of the image.

I am not checking the whole transformation, because it's a bit tedious to string check something that looks like this:
```html
<img decoding="async" src="https://res.cloudinary.com/canonical/image/fetch/q_auto,f_auto,w_650/https://ubuntu.com/wp-content/uploads/3b1d/dell-xps-2004.jpg" srcset="https://res.cloudinary.com/canonical/image/fetch/q_auto,f_auto,w_350/https://ubuntu.com/wp-content/uploads/3b1d/dell-xps-2004.jpg 350w, https://res.cloudinary.com/canonical/image/fetch/q_auto,f_auto,w_650/https://ubuntu.com/wp-content/uploads/3b1d/dell-xps-2004.jpg 650w, https://res.cloudinary.com/canonical/image/fetch/q_auto,f_auto,w_1300/https://ubuntu.com/wp-content/uploads/3b1d/dell-xps-2004.jpg 1300w, https://res.cloudinary.com/canonical/image/fetch/q_auto,f_auto,w_1950/https://ubuntu.com/wp-content/uploads/3b1d/dell-xps-2004.jpg 1950w" sizes="(max-width: 400px) 350w, 650px" loading="lazy">
```

If you guys think I should check the whole thing, let me know.

tests/test_blog_api.py
albertkol: > A question - does this only change images in the body of the article? What about the thumbnails on the index page?
> 
> I know that at the moment in e.g. `jp.ubuntu.com` the cloudinary prefix is added to thumbnails inside the template, but this isn't ideal. Really, if this module is going to do the transformation anywhere it should do it everywhere, and the prefix should be removed from templates.
> 
> That also therefore means that you need to make this at least another minor version change, as it effects the module's API.

This works now. Thumbnails are affected by the new `enable_cloudinary` parameter too.

albertkol: No idea why the reply was put under Joao's comment.

canonicalwebteam/blog/blog_api.py
albertkol: I created a second one `enable_cloudinary`. I can remove the `transform_links`, but I don't think it harms being there.

nottrobin: I don't feel strongly about the existence of this one, but I do generally consider unnecessary config options to be harmful. They are cognitive load for the developer to process, extra documentation for us to write, extra logic in the codebase...

nottrobin: Oh also, we're potentially confusing our own Devs, cos they might mistakenly think there are some cases where they shouldn't be doing this replacement, when in fact they should always do it.

albertkol: I added a new commit to remove `transform_links`. `enable_cloudinary` is the only new parameter. It controls the cloudinary prefix for both article images and thumbnails.

nottrobin:
@albertkol could you just remind me where we got to with displaying thumbnails / using the image_template module etc.? I think this might  be ready, but I forgot all the history =)

albertkol:
@nottrobin we are using `image_template` for both thumbnails and article images. For the thumbnails I am passing the url of the image with the dimentions 330x177 let it generate the rendered html image tag which I then echo on the site.

From what I was able to QA, it seems to be working..

canonicalwebteam/blog/blog_api.py
```diff
@@ -177,3 +202,33 @@ def _get_month_name(self, month_index):
...
+        soup = BeautifulSoup(content, "html.parser")
+        for image in soup.findAll("img"):
+            img_width = image.get("width")
+            img_height = image.get("height")
+
+            new_image = BeautifulSoup(
+                image_template(
+                    url=image.get("src"),
+                    alt="",
+                    width=img_width if img_width is not None else width,
```
nottrobin, line 225: couldn't this be simply:

```suggestion
                    width=img_width or width,
```

?

canonicalwebteam/blog/blog_api.py
```diff
@@ -12,11 +17,11 @@ def __init__(
         self,
         session,
         api_url="https://admin.insights.ubuntu.com/wp-json/wp/v2",
-        transform_links=True,
+        enable_image_template=True,
     ):
         super().__init__(session, api_url)
 
-        self.transform_links = transform_links
+        self.enable_image_template = enable_image_template
```
nottrobin, line 24: What do you think of `use_image_template`? Sounds slightly nicer to me.

PR #144, opened 2020-08-03
Fix 503 error for API consumers

### LINKED ISSUE #141: Returned 500 error when testing random endpoint
## Summary

`get_article` in `wordpress.py` is returning Not Found error (Line 110), for random enpoints like `/not-found` which is fine obviously because that post was not found. But when it is a a test that is doing that, should we return something like 404?

carkod:
## Summary
- Previously raised error returns a 503 to the API consumer
- This will ensure a 403 is returned and the consumer can handle this error
- This was also failing for some other projects in [Sentry](https://sentry.is.canonical.com/canonical/jp-ubuntu-com/issues/11286/?query=is%3Aunresolved)

## QA
- Go to [Kubeflownews](https://github.com/canonical-web-and-design/kubeflow-news.com) project and use this version
- `dotrun test` and make sure all tests pass now
- To do a contrast test, see that if you use the previous version, it will fail with a `503` error 

## Issues
Fixes #141

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,4 @@
+6.2.0: Fix 503 instead of 404 status for articles not found
 6.1.1: Restructure; Use `image_template` for images processing
 6.1.0: Restructure; Add API wrapper; Remove helpers
 6.0.0: Restructure; Force passing in session; Remove poetry
--- a/canonicalwebteam/blog/wordpress.py
+++ b/canonicalwebteam/blog/wordpress.py
@@ -111,9 +111,12 @@ class Wordpress:
         Get an article from Wordpress api
         :param slug: Article slug to fetch
         """
-        return self.get_first_item(
-            "posts", {"slug": slug, "tags": tags, "tags_exclude": tags_exclude}
-        )
+        try:
+            return self.get_first_item(
+                "posts", {"slug": slug, "tags": tags, "tags_exclude": tags_exclude}
+            )
+        except NotFoundError:
+            return {}
 
     def get_tag_by_id(self, id):
         return self.request(f"tags/{id}").json()
--- a/canonicalwebteam/blog/wordpress.py
+++ b/canonicalwebteam/blog/wordpress.py
@@ -113,7 +113,8 @@ class Wordpress:
         """
         try:
             return self.get_first_item(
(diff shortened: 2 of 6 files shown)


canonicalwebteam/blog/wordpress.py
```diff
@@ -1,3 +1,4 @@
+import flask
```
tbille, line 1: wordpress is a model, so framework agnostic.
This should be handled somewhere else.

CHANGELOG.txt
```diff
@@ -1,3 +1,4 @@
+6.2.0: Fix 503 error for API consumers
```
tbille, line 1:
```suggestion
6.2.0: Fix 503 instead of 404 status for articles not found
```
What do you think?

canonicalwebteam/blog/wordpress.py
carkod: Do you mean handling it in the `notFoundError` class ass an exception?

tbille: I mean that this class doesn't need to have a `flask` dependency. It's some kind of model, so should be as light as possible. This class just manages the requests to the wordpress API.

albertkol:
This change will fix my failing test from: https://github.com/canonical-web-and-design/kubeflow-news.com/pull/61
..so I QA-ed it with the help of that PR and it seems to work.

Code wise.. when releasing a new version, for as much as I know we also need to change the version number in `setup.py`.

carkod:
Thanks guys. Anymore comments? If not, please someone approve this

canonicalwebteam/blog/wordpress.py
```diff
@@ -111,9 +111,13 @@ def get_article(self, slug, tags=None, tags_exclude=None):
         Get an article from Wordpress api
         :param slug: Article slug to fetch
         """
-        return self.get_first_item(
-            "posts", {"slug": slug, "tags": tags, "tags_exclude": tags_exclude}
-        )
+        try:
```
jpmartinspt, line 114: Can we move the `try` block into the views and `flask.abort` with the appropriate codes?

jpmartinspt: Actually nevermind.

PR #148, opened 2020-08-19
Handle image_template exception

albertkol:
Catch `image-template` exceptions. If `image-template` cannot convert the image, skip it.

## Issue
Fixes #8087

## QA 
WIP


tbille:
@albertkol could you add a bit more details to why it wasn't working (like an example image)

albertkol:
> @albertkol could you add a bit more details to why it wasn't working (like an example image)

@tbille I tried not to spend too much time on it. It is weird. The `canonicalwebteam.image-template` is the package that applies the cloudinary prefix to the images. `image-template` is called by the blog module that is looking to apply the cloudinary service to all the images it finds. 

It looks like in some situations the images provided by the blog to the `image-template` have an invalid `src`. `image-template` has a condition where if the `src` is invalid it throws the exception `url must contain a hostname` which the blog modue was not catching therefore throwing a 500. So what I did is adding a `try - `catch` to handle it. If the `image-template` conversion fails, I just want to leave the image raw.

I checked some failing links after the fix on snapcraft.io and ubuntu.com but I cannot see any raw images left unconverted, so I am not quite sure what images with invalid `src` it was trying to convert.

tbille:
@albertkol This seems to be a bit of a "hacky" way of fixing this issue. We should dig a bit more in what is causing this error? Is it possible that the `image.get("src")` is fed with objects that are not images?

albertkol:
@tbille will dig more then and follow up with new findings

PR #150, opened 2020-09-18
Fix width image transform

albertkol:
Add a condition to check if the src of the image is relative. If it is we do not use the cloudinary service for it, since Cloudinary cannot reach that image.

## QA
Browse to https://ubuntu.com/blog/the-ubuntu-community-contributes-towards-saving-the-iberian-lynx
Get 500

Browse to: https://ubuntu-com-8318.demos.haus/blog/the-ubuntu-community-contributes-towards-saving-the-iberian-lynx
Get 200


carkod (approved):
LGTM

Issue #158, opened 2021-05-02
Going beyond existing pages causes 500


anthonydillon:
Go to:
- https://juju.is/blog?page=3
- https://ubuntu.com/blog?page=50000

Causes a 500 error but this error should be handled.

PR #166, opened 2023-10-25
Pass empty list if there is no category id

carkod:
WordPress API returns 400 if return list with empty string, because it treats it as a Truly parameter. Instead we should pass an empty list, which will be accepted as Falsy in the case there is no id for the category.

## QA

Go to https://ubuntu-com-13283.demos.haus/blog?page=2, check that it doesn't break anymore https://ubuntu.com/blog?page=2

Go to https://ubuntu-com-13283.demos.haus/blog?page=500, check that it return 404 now (exceeded max num of pages)

or 
- Pull down and checkout this PR
- Locally change `requirements.txt` of ubuntu.com project to: 
```
canonicalwebteam.blog @ git+https://github.com/canonical/canonicalwebteam.blog@fix-blog-paginated-page-wd-6988`
```
- Run `dotrun clean && dotrun`
- Go to http://0.0.0.0:8001/blog?page=2
- Go to http://0.0.0.0:8001/blog?page=500

## Issues
Fixes https://github.com/canonical/canonicalwebteam.blog/issues/158
Fixes https://warthogs.atlassian.net/browse/WD-6988

### DIFF
--- a/canonicalwebteam/blog/wordpress.py
+++ b/canonicalwebteam/blog/wordpress.py
@@ -1,4 +1,7 @@
+import flask
+
 from urllib.parse import urlencode
+from requests.exceptions import HTTPError
 
 
 class NotFoundError(Exception):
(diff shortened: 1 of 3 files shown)


canonicalwebteam/blog/wordpress.py
```diff
@@ -82,22 +85,29 @@ def get_articles(
...
+                    "categories": categories,
+                    "group": groups,
+                    "sticky": sticky,
+                    "before": before,
+                    "after": after,
+                    "author": author,
+                },
+            )
+        except HTTPError as error:
+            if error.response.status_code == 400:
```
akbarkz, line 106: what about other codes? will they just silently break?

carkod: Other 40X errors not likely to occur (e.g. if 403 happens then it should fail and not be handled because there must have been some change on server side that we need to update like tokens or secrets or firewall, or 402 similar error codes forbidden access are unusual and indicate something changed) this is a 400 bad request which is returned by the API endpoint and we need to handle as a result.

500 errors should break and they will show up in Sentry for us to fix and people should report it.

carkod: I would have wanted to even skip the `status_code` but then it fails because there is no json. So I wanted to be as specific as possible to avoid catching unwanted errors

akbarkz: okay, if you are sure that in all other error cases than 400 this will still work as expected, meaning it will report to Sentry and there won't be any weird behavior on the client side, then good to go

akbarkz (approved):

canonicalwebteam/blog/wordpress.py
carkod: Yes, in this specific case, we don't want it to break, because it's just somebody randomly entering a number that exceeds the limit. In theory WordPress API should return a 404, but for some reason they API specification doesn't determine that is how it should be.

jpmartinspt:
@carkod Ant merged something similar, might be worth rebasing and bumping to 6.4.2 if you want to further improve on this.

carkod:
> @carkod Ant merged something similar, might be worth rebasing and bumping to 6.4.2 if you want to further improve on this.

mm [latest version](https://pypi.org/project/canonicalwebteam.blog/) seems to be `6.4.1`?

carkod:
Ok looks like it's fixed in the latest version

PR #173, opened 2024-03-19
Remove args to see if gifs work

carkod:


bartaz:
Blog is throwing 500 on anything.

This older article has a gif: https://ubuntu-com-13680.demos.haus/blog/lunar-lobster-has-landed (https://ubuntu.com/blog/lunar-lobster-has-landed)

I don't think that just removing params would work. They are probably mandatory (in image template) but need another value?

We should test it out without blog first. Just using image template with some gif. Then figure out what params we should use for gifs.

bartaz:
There are 2 gifs on asset server for noble numbat, so we could try with them directly first:
https://assets.ubuntu.com/manager?q=noble&type=gif

Asset server even gives you "image tag" snippet (I don't know if these are the same options that are used in blog module):

```
          {{ image (
            url="https://assets.ubuntu.com/v1/31ae60ab-noble-numbat-mascot-animated.gif",
            alt="",
            width="600",
            height="600",
            hi_def=True,
            loading="auto|lazy"
            ) | safe
          }}
```

canonicalwebteam/blog/blog_api.py
```diff
@@ -252,9 +252,6 @@ def _apply_image_template(
                     alt="",
                     width=img_width or width,
                     height=img_height or height,
-                    hi_def=True,
```
bartaz, line 255: `hi_def` is mandatory, so it needs to be set to False if we want to try opposite value.

`e_sharpen` is not documented at all, so I don't know what it's doing or what value it has, but likely it's not mandatory (as we don't use it in other places it seems)

https://github.com/canonical/canonicalwebteam.image-template?tab=readme-ov-file#parameters

bartaz:
On top of the params passed here, there is a bunch of params our `image_template` passes to cloudinary by default, without any way to override them:

image


https://github.com/canonical/canonicalwebteam.image-template/blob/1aa6941e4f61cc96dbcf4b7a92d2c069304c8c89/canonicalwebteam/image_template/__init__.py#L34-L38

`f_auto` can potentially mean that gif is changed to some other format, possibly affecting animation quality

`q_auto` allows cloudinary to automatically choose the quality of transformed image, which can have impact on the gif quality as well

see: https://cloudinary.com/documentation/image_transformations

We would need to test out those options, one by one, to figure out how they affect gifs, and test them in different browsers (as they may affect result as well, because cloudinary adjusts format to what browser accepts).

Them quite likely we need to make some of the options set by default optional, so they can be overridden if needed. Or maybe we need to define a separate set of defaults for animated gifs.

PR #190, opened 2025-11-18
Validate image URLs before processing

Onibenjo:
## Done
Ensure image URLs have valid scheme and netloc before attempting to process them to prevent potential errors.
- Adds a URL validity check in _apply_image_template before calling image_template(...) .
- Skips any img with an invalid URL (non-HTTP(S) or missing host) to prevent errors from bad links.

Validating with `urlparse` ensures scheme is http / https and netloc is present, catching common invalid cases early and avoiding exceptions in image_template .

## QA

- Check out this blog post with invalid image URL: 
- https://canonical-com-2072.demos.haus/blog/everything-you-need-to-know-about-fips-140-3-on-ubuntu-videos
- Check that the page loads
- compare that to production
- https://canonical.com/blog/everything-you-need-to-know-about-fips-140-3-on-ubuntu-videos

## Ticket
https://warthogs.atlassian.net/browse/WD-31318

### DIFF
--- a/canonicalwebteam/blog/blog_api.py
+++ b/canonicalwebteam/blog/blog_api.py
@@ -261,7 +261,7 @@ class BlogAPI(Wordpress):
         for image in soup.findAll("img"):
             image_url = image.get("src")
 
-            if not image_url or "http" not in image_url:
+            if not image_url:
                 continue
 
             parsed = urlparse(image_url)
--- a/canonicalwebteam/blog/blog_api.py
+++ b/canonicalwebteam/blog/blog_api.py
@@ -3,6 +3,7 @@ import re
 import html
 from datetime import date
 from datetime import datetime
+from urllib.parse import urlparse
 
 # Packages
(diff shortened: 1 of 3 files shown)


canonicalwebteam/blog/blog_api.py
```diff
@@ -258,7 +259,13 @@ def _apply_image_template(
 
         soup = BeautifulSoup(content, "html.parser")
         for image in soup.findAll("img"):
-            if not image.get("src") or "http" not in image.get("src"):
+            image_url = image.get("src")
+
+            if not image_url or "http" not in image_url:
```
petesfrench: Do we need to check `"http" not in image_url` if we are also checking it below?

Onibenjo: Yes, you're right. It's a duplicate check.
I have removed the check

petesfrench (approved):
LGTM!

PR #200, opened 2026-01-23
chore: Add request timeout (30s), retry logic for transient 5xx errors

Onibenjo:
## Problem

The WordPress API client occasionally fails with ChunkedEncodingError: Response
   ended prematurely when fetching large payloads. This happens because:
  - No timeout configured, so requests can hang indefinitely
  - No retry mechanism for transient server errors
  - ChunkedEncodingError was unhandled, causing 500 errors

## Done
  - Add request timeout (30s) to prevent hanging connections on large payloads
  - Add retry logic (3 retries with exponential backoff) for transient 5xx errors
   (502, 503, 504)
  - Add error handler for ChunkedEncodingError that returns 502 Bad Gateway
  
  ## Issue
  https://canonical-ax.sentry.io/issues/88682441/?project=4510709886419024&query=is%3Aunresolved&referrer=issue-stream

https://warthogs.atlassian.net/browse/WD-33122
  
  ## QA
  Check that it loads okay 
ubuntu
  https://ubuntu-com-15976.demos.haus/blog
  https://ubuntu-com-15976.demos.haus/blog/feed
canonical
https://canonical-com-2200.demos.haus/blog

### DIFF
--- a/canonicalwebteam/blog/wordpress.py
+++ b/canonicalwebteam/blog/wordpress.py
@@ -1,17 +1,18 @@
-from .constants import (
-    CATEGORY_FIELDS,
-    TAG_FIELDS,
-    USER_FIELDS,
-    DEFAULT_POST_FIELDS,
-    POST_DETAILS_FIELDS,
-)
 import base64
 import re
-from urllib.parse import urlencode, unquote, quote
-from string import ascii_uppercase, ascii_lowercase
+from string import ascii_lowercase, ascii_uppercase
+from urllib.parse import quote, unquote, urlencode
 
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
+
+from .constants import (
+    CATEGORY_FIELDS,
+    DEFAULT_POST_FIELDS,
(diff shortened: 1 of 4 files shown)


canonicalwebteam/blog/wordpress.py
```diff
@@ -46,6 +51,16 @@ def __init__(
...
 
+        # Configure retry logic for transient failures
+        retry_strategy = Retry(
+            total=3,
+            backoff_factor=1,
+            status_forcelist=[502, 503, 504],
+        )
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        self.session.mount("https://", adapter)
+        self.session.mount("http://", adapter)
```
petesfrench: Why are we handling `http` connections? All connections going through the block module should we secure

petesfrench (approved):
LGTM

canonicalwebteam/blog/wordpress.py
Onibenjo: Good catch
Thanks for pointing

PR #219, opened 2026-07-02
fix: grant release-drafter permission to publish releases

Onibenjo:
Release Drafter has been failing to publish releases since the default `GITHUB_TOKEN` went read-only. The API call gets a 403 but the job still reports success, so nobody noticed and 7.1.0 never shipped: https://github.com/canonical/canonicalwebteam.blog/actions/runs/28586881424

This adds the `contents: write` permission the action needs, plus `pull-requests: read` for the release notes.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
+permissions:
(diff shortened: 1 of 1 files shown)


Ataku-B (approved):
LGTM!

